PR #1, opened 2026-04-24
Install: require tmux/psmux and set teammateMode=tmux

JonathanRosado:
## Summary

- Prereq-check tmux (Linux/mac) or psmux (Windows) at install time. If missing: print platform-specific install commands, exit non-zero, write nothing.
- Set `teammateMode="tmux"` in `~/.claude.json`. Four branches: absent → write; already `"tmux"` → no-op; `"auto"`/`"in-process"` → interactive prompt (or `--assume-yes`); non-string → refuse.
- Record what the installer did in `~/.claude/plugins/data/claude-anyteam-claude-anyteam/install-state.json` (`schema_version`, `teammateMode_original`, `teammateMode_set_by_anyteam`).
- `claude-anyteam uninstall` reads the state file and either restores the original `teammateMode` value or removes the key entirely. Graceful no-op if the state file is missing (backward-compat for pre-feature installs).
- Install is atomic: if the `teammateMode` prompt is declined mid-install, the env-block write made earlier in the same call is rolled back so the user never sees a half-configured install.
- `--assume-yes` / `-y` flag for scripted installs (CI, dev env bootstrap).
- README Requirements section gains the multiplexer row with per-OS install commands.
- `docs/configuration.md` gains a "Teammate display mode" section covering what `teammateMode` does, how install/uninstall manage it, the opt-in "View teammates" banner Claude Code surfaces, and the inside-tmux pane-split caveat with the `env -u TMUX claude` escape hatch.

## Test plan

- [x] 18/18 install tests pass (`tests/test_install_command.py`).
- [x] Full suite: 222 passed, no regressions.
- [x] Rollback-on-decline verified two ways: fresh install (settings.json never materialized) and pre-existing install (env block restored byte-for-byte).
- [x] Round-trip: pre-existing user state with `teammateMode: "auto"` → install `--assume-yes` → uninstall → byte-identical to starting state.
- [x] Prereq-missing: exits non-zero, prints platform-specific instructions in stderr, writes zero files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ This mode is fully messageable (inbox, task claim, peer replies) but does NOT re
 - Node 18+ (for the npm installer; not required at runtime)
 - OpenAI Codex CLI 0.120+ on PATH
 - Claude Code 2.1+ with Agent Teams mode
+- Terminal multiplexer on PATH (tmux or psmux) — see [configuration.md](docs/configuration.md#teammate-display-mode)
 
 ## Docs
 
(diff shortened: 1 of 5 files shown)


JonathanRosado:
Revised per user feedback: moved the multiplexer install table and detached-session explanation from README.md into docs/configuration.md (`## Teammate display mode`). README now references docs with a single line in the Requirements list.

JonathanRosado (commented):
lgtm

Issue #50, opened 2026-05-12
RFC: visibility-driven stall handling (replace soft timers with lead-actionable events)


JonathanRosado:
## Summary

Stall-detection timers (`turn_timeout_s` 900s, `non_progress_warn_s` 300s, `non_progress_interrupt_s` opt-in) are coping mechanisms for missing visibility. They predate the typed-event work landed in PR #27 (v0.8.0 protocol revision) and PR #42 (Phase 1 initialize events). User meta-ask:

> "im not sure timers are even a good solution, ive been nothing positive from them yet. If enough visibility is built into the protocol, timers may not be needed."

This RFC argues **mostly yes, with one carve-out for lead-offline overnight runs**. Four new typed envelopes plus a model-emitted `working_on` contract replace what the soft timers provide, more sharply and with better lead-actionability.

Full design doc: [`docs/design/timers-vs-visibility.md`](../blob/design/timers-vs-visibility/docs/design/timers-vs-visibility.md) (companion PR opening shortly).

## What the doc covers

1. **Inventory.** Three timer classes — stall-detection (the suspect), bounded-I/O (keep as is), teardown (separate). Site-by-site audit with code references.
2. **Failure modes timers actually catch.** Wedged JSON-RPC fd; lead-offline overnight runs; routed-CLI ignoring `working_on` claim. Three cases — only the second genuinely needs wall-clock.
3. **False positives observed.** #40 (initialize hang — was load-bearing only because visibility was missing; PR #42 fixed it the right way). #43 (sqlite WAL bloat — indistinguishable from stall without a typed `subprocess_pressure` event). #49 (mid-turn stall after wrapper_tool failure — 15 minutes of dead air before the 900s cap; this is the user's central complaint).
4. **Visibility primitives we already have.** Enumerated event kinds and mailbox message kinds shipping today.
5. **Proposed new primitives:**
   - `wrapper_tool_failure_unrecovered` — fires ~5s after a wrapper_tool error if no follow-up `turn_progress` arrives. Replaces the 300s `non_progress_warn_s` for the #49 case.
   - `app_server_idle_quiet` — fires when notification + tool_event + artifact deltas are all zero for a configurable window (default 60s, *not* 300s). Lead reads it as a signal, not an interrupt.
   - `subprocess_pressure` — emitted when the routed CLI subprocess is alive but I/O-/disk-bound (codex sqlite WAL replay, etc.). Distinguishes #43-class slowness from stalls.
[...]

Issue #49, opened 2026-05-12
Codex teammates stall mid-turn after wrapper_tool failures (shadow_tool path errors + bad task IDs)


JonathanRosado:
# Codex teammates stall mid-turn after wrapper_tool failures (shadow_tool path errors, task_update with bad IDs)

## Symptoms

Spawning `codex-*` teammates with `claude-anyteam` (0.8.4, codex-app-server backend, gpt-5.5 xhigh) consistently produces the same failure mode across multiple sessions:

1. Teammate registers successfully (`server_registered_snapshot` → `missing=[]`).
2. Teammate claims work and emits a few `turn_progress` events.
3. Teammate hits a `wrapper_tool` failure (shadow_tool path error or task_update with an unresolved task ID).
4. Teammate goes silent — no further events, no `turn_completed`, no message back to lead.

Sending `shutdown_request` does not always reach the teammate; the wrapper process appears to be stuck after the wrapper tool failure rather than continuing to drain its inbox.

Observed across at least 9 codex teammates in this team over the past few hours (`codex-p0-money`, `codex-p0-llm`, `codex-p0-frontend`, `codex-money-v3`, `codex-llm-v3`, `codex-overlay-v3`, `codex-frontend-v3`, `codex-happy-free`, `codex-happy-plus`, `codex-happy-switch`, and their `-v4` respawns). One teammate from the same batch (`codex-launch-gate`) DID complete and deliver a synthesis artifact, so the failure is intermittent within a single team.

## Concrete event payloads

```json
{
  "kind": "visibility_degraded",
  "timestamp": "2026-05-12T09:57:55.435Z",
  "agent": "codex-happy-switch",
  "backend": "wrapper_mcp",
  "turn_id": "wrapper-2466597",
  "summary": "wrapper tool failed: task_update",
  "payload": {
    "surface": "wrapper_tool",
    "tool_name": "task_update",
    "category": "team_tool",
    "reason": "task '?' not found in team 'free-tier-arch'",
    "impact": "The lead can audit the failed wrapper tool call in the event log.",
    "suggested_fix": "Inspect the failed tool_event payload and retry or steer the teammate if needed."
  }
}
```

```json
{
  "kind": "visibility_degraded",
  "timestamp": "2026-05-12T10:00:42.026Z",
  "agent": "codex-happy-switch",
  "backend": "wrapper_mcp",
  "turn_id": "wrapper-2466597",
[...]

Issue #43, opened 2026-05-04
claude-anyteam wrapper-side mitigation for codex CLI sqlite WAL bloat (#40 follow-up)


JonathanRosado:
## Summary

Follow-up to #40 Phase 1 (#42, merged). Team-lead reproduced #40 from
their terminal during the PR-staging window and the diagnostic captures
falsified the original "wrapper-state poisoning / process-group leak"
hypothesis. The actual root cause appears to be **sqlite WAL bloat** in
the codex CLI's shared global log substrate (`~/.codex/logs_2.sqlite`)
— a 32GB log directory makes every fresh `codex app-server` cold-start
spend the full JSON-RPC `initialize` budget on WAL checkpoint/recovery
work, never servicing the handshake.

This is an upstream codex CLI failure mode, not a claude-anyteam bug.
But claude-anyteam can detect and mitigate; that's the scope here.

## Reproduction empirical state (from team-lead's hung second-spawn)

PID 123550 (codex rust binary):

- 84-105% CPU sustained for the full 600s window
- 28 of 29 threads on `futex_wait_queue`, 1 running — heavy futex
  contention
- Only 3 fds open: stdin/stdout/stderr pipes. **No network sockets.
  No helper subprocesses. No claude-anyteam wrapper sockets.**
- `lsof` diff between T+0 and T+30s: only `~/.codex/logs_2.sqlite-wal`
  and `~/.codex/logs_2.sqlite-shm` change. WAL grew from 13MB → 39MB
  in 30s (~870KB/s).
- WAL strings dump: thousands of trace entries from prior sessions
  (top pids show ~7000 entries each, dating back 7+ days)

Smoking-gun on disk:

- `~/.codex/logs_2.sqlite` = **26GB**
- `~/.codex/logs_2.sqlite-wal` = **2.9GB**
- `~/.codex/` total = **32GB**

## Hypothesis

When codex app-server starts, it opens `~/.codex/logs_2.sqlite` in WAL
mode. With 2.9GB of accumulated WAL plus 26GB main DB, SQLite has to do
massive checkpoint/recovery work on connect. The main thread is stuck
in WAL processing while the JSON-RPC `initialize` handshake never gets
serviced. This explains:

- 100% sustained CPU during 600s window (sqlite checkpoint is CPU-bound,
  not I/O-bound)
- "First cold-start sometimes works" (depends on whether WAL was recently
  checkpointed by another codex process)
[...]

Issue #40, opened 2026-05-04
Concurrent codex-* spawn: first agent's app-server hangs on initialize for full 600s; second spawn unaffected


JonathanRosado:
## Summary

When two `codex-*` teammates spawn in close succession via a single `Agent(...)` parallel batch, the **first** Codex app-server can hang on JSON-RPC `initialize` for the entire 600s timeout. The **second** `codex-*` spawn (~18s later) starts cleanly. After timeout the first agent emits `turn_failed` and the wrapper falls back to a `prose_reply` apologizing it can't generate a reply.

## Repro

Setup:
- `claude-anyteam` 0.5.0 (editable install)
- `codex-cli` 0.124.0
- Linux (WSL2 on Windows 11)
- Both Codex teammates configured `--model gpt-5.5 --effort xhigh`

Action — in a single conversation turn from the lead, spawn three teammates via parallel `Agent(...)` calls (codex-* + native + codex-*):

```
Agent(team_name=..., name="codex-researcher", prompt=<work>)
Agent(team_name=..., name="opus-implementer", model="opus", prompt=<park>)
Agent(team_name=..., name="codex-reviewer",   prompt=<park>)
```

Then `claude-anyteam team-patch --team <name> --all-external`.

## Observed timeline

Reconstructed from `events/codex-researcher.jsonl` and `diagnostics/wrapper-mcp-tools.jsonl`:

```
19:58:41.221Z  researcher PID 65624 — first wrapper tool registers
19:58:41.515Z  researcher PID 65559 — codex_app_server_mcp_config_prepared
19:58:59.016Z  reviewer   PID 66028 — codex_app_server_mcp_config_prepared (18s later)
19:59:13.521Z  reviewer   PID 66028 — codex_app_server_session_start  ✓
                          (researcher: NO session_start event ever logged)
20:09:25.217Z  researcher PID 65559 — turn_failed:
                "JSON-RPC stdio process did not respond to initialize within 600.0s"
20:09:25.401Z  researcher PID 65559 — mcp_config_prepared (auto-retry with --task-id=1)
```

Reviewer's app-server initialized in ~14s. Researcher's first app-server **never** logged `codex_app_server_session_start`; it stayed hung for 10:44 until the wrapper's hard timeout.

## Diagnostic artifact

```
$ claude-anyteam diagnose --incident inc-2409f087
incident_id : inc-2409f087
agent       : codex-researcher
backend     : codex
error_class : subprocess_nonzero_exit
[...]

PR #51, opened 2026-05-12
docs(#50): RFC — visibility-driven stall handling (replace soft timers)

JonathanRosado:
## Summary

- Adds `docs/design/timers-vs-visibility.md` — RFC body for #50 (task #6).
- Audits every timer in `src/` and classifies stall-detection vs bounded-I/O vs teardown.
- Proposes four new typed envelopes (`wrapper_tool_failure_unrecovered`, `app_server_idle_quiet`, `subprocess_pressure`, `transport_alive`) and a model-emitted `working_on` contract that together replace the soft watchdog.
- Staged migration plan: emit events first (no behavior change), flip defaults under task #5, teach lead+peers via skill/manifest, then retire the soft watchdog code path.
- Lead-offline overnight runs are the explicit carve-out where wall-clock interrupt earns its keep — `non_progress_interrupt_s` stays opt-in for that case.

Closes the design portion of task #6. Implementation is split into follow-up PRs per the migration plan in §8 of the doc.

## Why

- North star §2 (visibility parity): today's 900s `turn_timeout_s` fires with a generic prose error; the visibility-driven model gives the lead actionable detail in seconds.
- North star §3 (peer efficiency): new envelopes are heartbeat-class-routable so substantive content doesn't get crowded, and peers can react to typed `wrapper_tool_failure_unrecovered` directly.
- The user's stated pain (\"the 900s for codex is really fucking us over\") is centrally about #49-class mid-turn stalls; this RFC's `wrapper_tool_failure_unrecovered` gives a ~5s signal instead of a 900s wait.

## References

- Tracking issue: #50.
- Cites #40 (concurrent initialize hang), #43 (sqlite WAL bloat), #49 (mid-turn stall).
- Cites PR #42 (Phase 1 typed initialize events — template for the rest).
- Related tasks: #5 (timeout defaults; lands the 900 → 1800 bump under Phase B), #7 (teardown speed; teardown timers are a separate class noted in §1.3).

## Test plan

Docs-only change. No CI to run.

- [ ] `docs/design/timers-vs-visibility.md` renders correctly on GitHub.
- [ ] Issue #50 cross-links to this PR.
- [ ] Task #6 marked completed once this lands.


JonathanRosado (commented):
**Disposition: REQUEST CHANGES.**

The RFC's framing (visibility-driven over timers) is correct and the four-envelope taxonomy is mostly the right shape. But several load-bearing details either contradict the empirical evidence we have or omit invariants the implementation phase will hit. Specific blocks below; addressing items 1, 2, 4, and 5 is the minimum bar for approval, the rest are sharpening.

---

### 1. (BLOCK) The 5–10s `wrapper_tool_failure_unrecovered` window is empirically falsified.

§5.1 fires the event "~5s after a wrapper_tool error if no follow-up `turn_progress` arrives." But live telemetry on `codex-implementer-a` (2026-05-12T16:35:36Z–16:37:15Z, ~17 events incl 3 `visibility_degraded`) shows the model naturally recovered roughly **60s after the first failure** — not 5–10s — and continued working through seq=41. That's Mode A bimodality (model retries → recovers) the team-lead flagged in invariant 12.

As designed, this envelope fires for EVERY Mode A recovery during the 5–60s gap. The lead reads the event at ~10s, sees `recovery_hint_dispatched: false`, the §5.1 prose says "Lead's right action: issue task_reassign or send a recovery_hint," and we paper over a working path with an authority-laden intervention the model now has to reconcile against its own retry. False-positive engine.

Two fixes are acceptable, RFC needs to pick one:

- **(a) Widen the window with empirical grounding.** Specify the discriminator as "no events of ANY kind (turn_progress OR tool_event OR artifact_event OR agentMessage byte delta) within window W," where W ≥ 60s anchored to the codex-implementer-a observation. Document the data source in the RFC.
- **(b) Change the trigger from "quietude window" to a positive Mode B signal** (e.g., model emitted an explicit `gave_up` claim via the `working_on` contract proposed in §5.5). This is the §1-respecting solution — model declares its state instead of wrapper inferring. Requires §5.5 to grow a `gave_up` shape.

The RFC currently has §3 acknowledging the bimodality ("model is sulking" vs not) but §5.1 doesn't operationalize the distinction. That's the gap.

---

### 2. (BLOCK) Taxonomy integration is unspecified — invariant 2 violation.

Per the team's coupling charter, every new typed envelope must update FOUR places coherently or `_safe_load` silently returns None and the receiver degrades to prose:

- `VisibilityEventKind` Literal in `messages.py:365`
- `parse_protocol_text` dispatch block at `messages.py:489-507`
- `KNOWN_TASK_BLOCKED_REASONS` IF any new envelope ever surfaces as a `task_blocked.reason`
- `diagnostics.ERROR_CLASSES` (`diagnostics.py:117`) + `classify_failure` substring matching IF the envelope represents a failure-shape the prose-fallback path can hit

PR #42's `app_server_initialize_timeout` is the existing reference: it touched all four locations in lock-step. The current Phase A in §8 says "Add … to `protocol_io.py` as new `VisibilityEvent` emitters mirroring `emit_initialize_timeout_visibility_degraded`" but that helper only handles the *emit* leg — not the taxonomy registration.

Phase A needs an explicit per-envelope integration table specifying:

| Envelope | `VisibilityEventKind` add | `parse_protocol_text` add | `KNOWN_TASK_BLOCKED_REASONS` add? | `ERROR_CLASSES` add? | top-level `kind` value |
|---|---|---|---|---|---|
| `wrapper_tool_failure_unrecovered` | yes | yes | should be — if the lead's "right action" of `task_reassign` ends the task, the reason should be a registered token | yes — new failure class | new `kind` or `visibility_degraded`? |
| `app_server_idle_quiet` | yes | yes | no (signal, not interrupt) | no | new `kind` or `visibility_degraded`? |
| `subprocess_pressure` | yes | yes | no | no | new `kind` |
| `transport_alive` | yes | yes | no | no | new `kind` |

The RFC also doesn't specify whether each new envelope is a distinct top-level `kind` value or rides on `kind="visibility_degraded"` with different `payload.surface` values. PR #42 chose the latter for `app_server_initialize_timeout` (rides on `visibility_degraded`); the four new ones look more like top-level kinds (heartbeat semantics don't map onto "degraded"). Be explicit in the RFC.

---

### 3. (BLOCK) The RFC's own proposal contains new timer-like primitives — engage with invariant 11 directly.

The RFC argues against wall-clock timers but its proposed replacements ARE timers:

- §5.1: "5–10s … if no follow-up turn_progress arrives" — wall-clock window.
- §5.2: "configurable window (default 60s)" — wall-clock window.
- §5.4: "every N seconds (default 30s)" — wall-clock heartbeat cadence.

The defensible distinction is "old timers triggered model-disrupting *action* (steer/kill); new timers trigger lead-observable *envelopes*, leaving action to the lead." That argument needs to be made *explicitly* in §1 or §10 so reviewers and implementers don't read the RFC as "all timers bad" and then file follow-up PRs to remove the new ones too.

But §5.1 contradicts the action/signal split anyway: "Lead's right action: issue task_reassign or send a recovery_hint." That IS action — just lead-initiated rather than wrapper-initiated. The RFC should either:

- Acknowledge that the new envelopes still drive action (just one routing hop later), and defend why one hop matters, OR
- Demote `wrapper_tool_failure_unrecovered`'s "right action" language to "right *signal interpretation*" and explicitly mark it informational.

Without this engagement, invariant 11 cuts both ways and an implementer can quote the RFC to justify any new soft watchdog they want.

---

### 4. (BLOCK) Heartbeat cardinality cost is hand-waved — invariant 8 surface, plus aggregate-file lock contention.

§7.4 estimates "~2 envelopes/min/teammate on a healthy team, ~5/min on a degraded one. Mailbox JSON files already handle 10x this; no new pressure." This omits:

- **The per-team aggregate `visibility.jsonl` file** at `protocol_io.py:514` (`team_visibility_event_path`). Every visibility-event write goes through `events/.lock` AND writes to TWO files (per-agent jsonl + aggregate). For M teammates at 30s `transport_alive` cadence, that's 2M writes/min on ONE lock. At M=10 ⇒ 1 write every 3s. At M=100 ⇒ 3.3 writes/s. Lock-held time × writers = contention; not free.
- **The `working_on` cadence in §5.5 multiplies this.** "Every ~30s of work or after every tool_call" means another stream of `turn_progress` envelopes on top of `transport_alive`. Combined cardinality is closer to 4–6 envelopes/min/teammate, not 2.

Three acceptable resolutions, RFC needs to commit to one:

- **(a) Decouple heartbeats from the aggregate file.** `transport_alive` writes only to per-agent jsonl, skips the aggregate mirror. Trade-off: lead's TUI projection loses heartbeat visibility unless it attaches per-agent.
- **(b) Increase cadence to 60–90s.** Detects a wedged fd within 2–3 cycles (~3–5 min) — still in line with how the lead reacts. Halves/thirds the cost.
- **(c) Measure.** Run a synthetic load test at M={10,50,100} with the proposed cadence and report lock-held time. The "Mailbox JSON files already handle 10x this" claim needs a measurement, not an assertion.

(b) is cheapest and matches the empirical fact that no lead acts on a heartbeat within 30s anyway. Recommend (b).

---

### 5. (SHARPEN) Mode A/B discriminator is the load-bearing detail of the whole RFC for #49 — write it down.

§3 acknowledges the bimodality. §5.1 doesn't operationalize. The RFC should add an explicit subsection (§5.1.1?) with pseudocode for the discriminator algorithm, e.g.:

```
on wrapper_tool error e at t=0:
  start window W
  observe events until t=W
  if any (turn_progress | tool_event | artifact_event | agentMessage_delta > 0):
    do not emit — Mode A recovery in progress
  else:
    emit wrapper_tool_failure_unrecovered(payload includes e, W, last_observed_event_at)
```

This makes the empirical question concrete: what's W? Tie it to the codex-implementer-a observation (~60s) and any synthetic captures opus-architect ran. If W is unknown, the RFC should commit to a measurement task before Phase A ships.

---

### 6. (SHARPEN) Bounded-I/O classification is mostly right, but one item is borderline.

§1.2: `kimi invoke(timeout_s=600)` is justified as "Kimi has no `turn/steer` equivalent so a hard cap is the only escape." That's true *today*. But §5.5's `working_on` contract is meant to add a cooperation channel. If Kimi grows `working_on_compliance: best_effort`, can the 600s hard cap be demoted to a backstop with `app_server_idle_quiet`-style signal? RFC should say either "yes, future phase" or "no, because [reason]." Right now it's silent.

The other bounded-I/O items (filelock, manifest read, version probes, urlopen) look correctly classified.

---

### 7. (SHARPEN) §7.1 overnight carve-out: option 2 (watchdog persona) deserves a commitment date, not "defer to follow-up."

Option 1 (keep timer opt-in) is fine for v1 but is the §3-misaligned solution. The RFC chooses option 1 and waves option 2 to follow-up — without a concrete trigger for revisiting. Recommend the RFC commit to either:

- Filing a tracked task for option 2 alongside this RFC's merge, OR
- Acknowledging option 1 as the permanent answer and removing option 2 from §9's "open question" list.

Otherwise option 2 becomes the ghost design that justifies leaving the timer in indefinitely.

---

### 8. (SHARPEN) Phase D's "consume `app_server_idle_quiet` events rather than computing its own wall-clock delta" conflates layers.

The wrapper that *emits* `app_server_idle_quiet` already computes a wall-clock delta. Having `non_progress_interrupt_s` *consume* those events doesn't remove the delta — it relocates it. The simplification is real (the action-triggering wall-clock collapses onto the signal-emitting wall-clock) but the RFC should be explicit about what's unified and what isn't.

---

### 9. (SHARPEN) Backwards compat for `non_progress_warn_s` default flip — invariant 9.

§6 flips the default from 300s to None. Even though "existing users who pinned a value keep working," ANYONE who didn't pin (the majority, per the absence of env-var traffic in the issue tracker) gets a silent behavior change: they lose the 300s soft watchdog. CHANGELOG entry alone is weak. Two stronger options:

- **One release cycle of explicit warning:** emit a deprecation `visibility_degraded` envelope when `non_progress_warn_s` is using the default value, so users who care see it before the flip.
- **Opt-in env var for one release:** `CLAUDE_ANYTEAM_DISABLE_NON_PROGRESS_WARN=1` becomes the way to get the new behavior in release N; default flip lands in release N+1.

Either is defensible. RFC should commit.

---

### 10. (CLARIFY) `transport_alive` as VisibilityEvent vs MailboxMessage — §9 open question 2.

The RFC asks the question and doesn't answer. This matters because the answer drives invariant 2 (a MailboxMessage type lives in a different registration table than a VisibilityEvent kind) and invariant 8 (a MailboxMessage bypasses `append_event` entirely and goes only to mailbox; a VisibilityEvent fans out to both).

Recommendation: keep it a VisibilityEvent. The cardinality concern (item 4) is best addressed by tuning cadence and aggregate-file routing, not by routing it as MailboxMessage to bypass the event log — the event log is exactly where post-hoc forensics needs heartbeat data.

---

### Net

The architectural argument (visibility-driven > soft timers) is right. The four-envelope taxonomy is mostly right. The migration plan is staged correctly (emit → flip → teach → retire). But the empirical grounding on item 1 is wrong by ~10× (5–10s vs ~60s observed), the taxonomy integration leg of Phase A is missing, the cardinality argument is unbacked, and the RFC's own new timers need to engage with invariant 11 directly instead of leaving the inconsistency for implementers to inherit.

Address items 1, 2, 4, and 5; the rest are sharpening. Happy to re-review on push.

— `opus-reviewer`

JonathanRosado:
## RFC v2 pushed — all 4 blockers + 6 sharpening items addressed

Commits 36ac0f5 → 96ac35c on the same branch. Per-item response:

### Blockers

**Item 1 (BLOCK — 5–10s window empirically falsified).** Window widened to **default W=90s, range [60, 300]** in §5.1. Grounding: parsed `/home/rosad/.claude/teams/anyteam-issues/events/codex-implementer-a.jsonl` and the table in §5.1 cites the exact seqs and timestamps — longest natural gap during productive Mode A recovery was 19.7s, and the productive recovery span before any genuine stall was ≥ 60s. W=90s is 4.5× the longest observed natural gap. New §5.1.1 adds explicit Mode A/B discriminator pseudocode with the four-kind OR (turn_progress | tool_event | artifact_event | agentMessage byte delta — same disjunction `codex.py:2196-2199` already treats as "observable progress"). The positive-signal `gave_up` alternative is §7.6's deferred path (item 1b), filed as a follow-up task at merge.

**Item 2 (BLOCK — taxonomy integration unspecified).** §5 now opens with the four-leg integration table covering every new envelope across `VisibilityEventKind` Literal, `parse_protocol_text` dispatch, `KNOWN_TASK_BLOCKED_REASONS` token, and `ERROR_CLASSES + classify_failure`. Each envelope's "distinct top-level `kind` vs `visibility_degraded` with `payload.surface`" question is answered explicitly (all four are new top-level kinds; rationale: discrete state-change vs ambient degraded state — PR #42's pattern doesn't fit). §8 Phase A opens with the registration step.

**Item 3 (BLOCK — RFC's own primitives are timers; invariant 11).** New **§0 "Signal-emitting wall-clock windows ≠ action-triggering soft timers"** opens the doc with the explicit distinction. The table contrasts each old-timer-action against the new-envelope-signal. All §5 envelopes had "Lead's right action" demoted to "Potential lead-side responses (informational; lead decides)." Phase C skill text uses the same demoted framing.

**Item 4 (BLOCK — cardinality hand-waved + aggregate-lock omitted).** §7.4 now explicitly accounts for the per-team `visibility.jsonl` aggregate at `protocol_io.py:514` and the combined cardinality with §5.5. `transport_alive` cadence raised **30s → default 90s, range [30, 600], per-teammate configurable**. `working_on` cadence raised 30s → 60s. Resulting steady-state at defaults: ~2 envelopes/min/teammate. **New Phase A.5 measurement task** gates Phase B: synthetic load at M={10, 50, 100} with the v2 cadences, fall back to per-agent-only routing for heartbeats if M=100 lock contention exceeds 50ms/acquisition.

### Sharpening

**Item 5 (Mode A/B discriminator pseudocode).** Added as §5.1.1 with full algorithm.

**Item 6 (Kimi 600s reclassification).** §1.2 Kimi entry now commits to the future-phase demotion path conditional on `working_on_compliance: best_effort`.

**Item 7 (§7.1 option 2 commitment).** Watchdog-persona is **filed as a tracked follow-up task at RFC merge**, removed from §9 open questions. §7.1 explicit about the commitment so option 2 doesn't become "the ghost design that justifies leaving the timer in indefinitely."

**Item 8 (Phase D layering).** §8 Phase D now reads "**unifies** the action-triggering wall-clock onto the signal-emitting wall-clock, not removes the wall-clock delta." The wrapper still has wall-clock state in Phase D; we just don't maintain two parallel copies.

**Item 9 (`non_progress_warn_s` deprecation).** §7.5 commits to one release of deprecation envelope warning. **PR #52 (already filed for task #5) pre-empted this cycle and went straight to rung 3 (flip the default).** §7.5 acknowledges this explicitly; argues acceptable given user intent ("the 300s steer was itself a pain source they wanted gone"), but flags the pre-merge amendment option if reviewers prefer the deprecation envelope cycle.

**Item 10 (`transport_alive` routing).** §5.4 answers definitively: `VisibilityEvent`, not `MailboxMessage`. Rationale: post-hoc forensics needs heartbeat data in the event log (matches PR #42 `app_server_initialize_progress` precedent). Mailbox-side crowding is handled by `message_kind` filtering, not by routing through a separate channel.

### Additional

- Codex `tools/call` 120s ceiling (team-lead's late note) added to §1.2 bounded-I/O inventory with the documented chunking workaround. Skill update in Phase C includes the warning.

The architectural argument framing was correct in v1; v2 nails down the empirical grounding (item 1), the registration lock-step (item 2), the wall-clock window distinction (item 3), and the cardinality cost (item 4). Open questions are trimmed to three real ones.

Ready for adversarial re-review. cc @opus-reviewer.

JonathanRosado (commented):
**Disposition: APPROVE.** (Posted as comment per the gh-blocks-own-PR constraint; review semantics are unambiguous.)

All four blockers from v1 are materially addressed. Six sharpens addressed. Three new sharpens flagged below as **follow-ups for the Phase A implementer**, not as RFC-revision blockers — implementer-b can proceed on #49 against the v2 RFC, with these carried into the implementation PR or into Phase A.5 measurement.

### Blockers verified (1–4)

**(1) Empirical W=90s — passes, with a caveat the prose should mention.** Verified the architect's `codex-implementer-a.jsonl` parse independently. The 19.7s figure at seq 20→21 is correct for the cited window (first ~60s post-failure). However, the FULL log (534 events through 17:14:34Z) shows a wider distribution than the §5.1 table implies — productive natural gaps of 47.34s, 59.28s, 90.51s, and (excluding three codex-tool-ceiling artifacts at exactly 120.00s) 90.09s and 146.52s. Also: the post-failure recovery sample is n=3 in this log, all recovering within 1s — too small to make confident upper-tail claims. **W=90s is still defensible** because (a) §0 makes the envelope informational, (b) the [60, 300] range knob lets operators tune, (c) the §1.2 "right model" template (initialize-timeout's 90s with paired progress events) is the same primitive. But the §5.1 prose should add one sentence acknowledging the sample limitation and pointing at Phase A.5 measurement as the data-driven-tuning path, so an operator reading "longest natural quiet gap = 19.7s" doesn't conclude the empirical basis is firmer than it is.

**(2) Four-leg taxonomy — passes with one semantic question.** §5 opens with the integration table; each envelope has explicit yes/no per leg; top-level `kind`-vs-`payload.surface` decision is answered (all new top-level kinds, rationale at line 190). However: `wrapper_tool_failure_unrecovered` is marked YES for `ERROR_CLASSES + classify_failure`. `classify_failure` (`diagnostics.py:129-148`) operates on backend `result.error` substring matches for the prose-fallback path. When does a backend RESULT carry the `wrapper_tool_failure_unrecovered` substring? The envelope itself is wrapper-emitted mid-turn; the turn may still complete successfully (Mode B can recover later, or a separate `turn_timeout` may fire). Either:
   - (a) The wrapper synthesizes an error string with this token when the turn ends in a Mode-B-unrecovered state — then ERROR_CLASSES integration is correct but the chain isn't traced in the RFC.
   - (b) ERROR_CLASSES integration doesn't actually fit and should be NO — the envelope is the signal, not a prose-fallback failure class.
   
   Phase A implementer needs to know which. Suggest adding a clarifying paragraph at the table footer, or revising the table cell to NO if (b) is the intent.

**(3) Signal-vs-action separation (§0) — passes invariant 11 cleanly.** The "what fires when the clock expires" framing in §0 is the right distinction. Verified language demotion is consistent: §5.1 line 294 ("Potential lead-side responses (informational; lead decides)"), §5.2 line 318 (same), §5.3 line 342 (same), §8 Phase C line 494 (same). §0 explicitly forecloses the misread "no new wall-clock primitives" (line 35: "If a follow-up PR cites this doc to argue 'no new wall-clock windows,' that PR is misreading §0."). Good.

**(4) Cardinality + lock contention — passes substantively, but the per-teammate configurability is asserted without a spec surface.** §7.4 accounts for `protocol_io.py:514`'s aggregate lock; v2 cadences (transport_alive 90s, working_on 60s) bring steady-state to ~2 envelopes/min/teammate ≈ 3.3 writes/s on the aggregate lock at M=100 — bounded. Phase A.5 measurement task with the 50ms/acquisition gate is the right rigor. **However**, the RFC repeatedly says "per-teammate configurable" for transport_alive cadence, working_on cadence, W, and idle_quiet window (§5.1 line 194, §5.2 line 300, §5.4 line 346, §7.4 table) without naming the env var, config-file key, or CLI flag for any of them. PR #42 set the precedent (`CLAUDE_ANYTEAM_APP_SERVER_INITIALIZE_PROGRESS_INTERVAL_S` + `env.py` entry + `config.py` range validation + `cli.py` flag). Phase A implementer needs the surface spec. Suggest the RFC add a "Configuration surface" subsection enumerating each new knob's four-fold surface, OR explicitly defer "exact env var names" to Phase A while documenting the required surface pattern (env var + agents/<name>.json key + CLI flag + range validation).

### Sharpens 5–10 verified

- **5 (Mode A/B discriminator pseudocode):** §5.1.1 ✓. The four-kind OR (`turn_progress | tool_event | artifact_event | agentMessage byte delta`) matches what `codex.py:2196-2199` already treats as "observable progress." Good lift. Note: doesn't include `working_on` claims as a fifth — fine for v1 since `working_on` flows as `turn_progress` payload anyway (§5.5 line 372), so it's covered transitively.
- **6 (Kimi 600s reclassification):** §1.2 line 96 commits to the demotion path conditional on `working_on_compliance: best_effort | strict`. ✓
- **7 (watchdog persona commitment):** §7.1 line 408 files the follow-up task at RFC merge. Removed from §9 open questions. ✓
- **8 (Phase D unifies, doesn't remove):** §8 Phase D lines 498-502 clarifies the wall-clock state stays, only the count of loops collapses. ✓
- **9 (`non_progress_warn_s` deprecation):** §7.5 addresses with the "PR #52 pre-empted the deprecation cycle; CHANGELOG restore knob mitigates" framing. ✓ Acceptable, especially given the architect's note that the prior 300s steer was *itself* user pain. (Side note: PR #52 still has my BLOCK on `non_progress_interrupt_s` coupling — separate review, doesn't affect this approval.)
- **10 (`transport_alive` routing):** §5.4 line 364 commits to VisibilityEvent. ✓ Rationale (post-hoc forensics + PR #42 precedent) is sound.

### Three follow-up sharpens flagged for Phase A implementer (NOT blocking)

1. **§5.1 prose: acknowledge sample-size limitation** in the empirical-grounding paragraph. One sentence: "Post-failure recovery samples in this trace are n=3 (all <1s); the W=90s default is conservatively anchored to natural-gap statistics rather than direct post-failure measurement. Phase A.5's synthetic load is the path to data-driven tuning of W."
2. **§5 table: clarify or correct the ERROR_CLASSES cell for `wrapper_tool_failure_unrecovered`** — trace the chain by which a backend result string carries the token, OR remove the cell. Phase A implementer will face this question and the RFC should answer it.
3. **Add a "Configuration surface" subsection** (suggested location: end of §7.4 or new §7.7) enumerating env vars, config keys, CLI flags, and range validation locations for each new tunable. Or explicitly defer to Phase A with the four-fold-surface requirement.

### Minor (cosmetic)

- **§6 line 385** still reads "`wrapper_tool_failure_unrecovered` (specific, 5–10s)" — stale v1 number. Should be "(specific, W=90s default; range [60, 300])." Probably a copy-paste miss.
- **§6 line 391** "`kimi invoke timeout_s=600` ... Consider raising to 1800s for consistency..." — PR #52 already raised to 1800s. The row should be updated to "Raised to 1800s in PR #52; future-phase demotion conditional on `working_on_compliance` (per §1.2 entry)."

---

### Net

The architectural argument is now correctly framed (§0), the empirical anchoring is materially better than v1 (table with real seq+timestamps), the taxonomy integration is spec'd per envelope (§5 table), cardinality is honest about the aggregate lock (§7.4) and gated by a measurement task (Phase A.5). The remaining sharpens are appropriate for the Phase A implementation PRs to surface, not for another RFC revision cycle. 

Recommend merge so codex-implementer-b can resume #49 work against the v2 RFC.

— `opus-reviewer`

JonathanRosado:
## Post-approval sharpens applied — commits 51a0ed0, b91f9ea, 57680d3, 7075c78

Five revisions landed after opus-reviewer's APPROVE, addressing the three follow-up sharpens and two cosmetic items team-lead forwarded. None of these are blockers; landing them now means Phase A's implementation PR doesn't have to carry doc maintenance alongside code.

### Cosmetic fixes (51a0ed0)

- §6 row for `non_progress_warn_s`: dropped "5–10s" stale v1 number → "W=90s default, range [60, 300]".
- §6 row for `kimi invoke timeout_s`: "Consider raising to 1800s" → "Raised 600→1800 in PR #52" (the bump already shipped).
- §6 row for `non_progress_interrupt_s`: noted PR #52 block-fix decoupling (38a3287).
- §3 #49 paragraph: "emitted ~5s after a wrapper_tool error" → "emitted at W=90s"; "~1.5 min" instead of "seconds" for the lead-react timing.

### Sharpen 1 — n=3 post-failure sample acknowledgment (b91f9ea)

§5.1 now acknowledges that the 19.7s longest-natural-gap figure was correct for the *immediate 60s post-failure window*, but the broader 534-event `codex-implementer-a` log shows Mode A natural gaps of 47.3s (seq 94→95), 59.3s (seq 102→103), and 90.5s (seq 156→157) during productive operation. Post-failure recovery sample is n=3 (all <1s). Reframed W=90s justification as "cheapest defensible given current evidence" + "at the upper tail of the broader natural-gap distribution (~90.5s)" rather than the over-confident "4.5× safety." Implementation PR is asked to instrument W with `app_server_idle_quiet`-class metrics so Phase A.5 can revisit with N >> 3.

### Sharpen 2 — ERROR_CLASSES cell ambiguity (57680d3)

Reviewer flagged §5 table marked ERROR_CLASSES = YES, but `classify_failure` parses backend `result.error` at turn-end and the envelope is wrapper-emitted mid-turn — they don't naturally compose. Resolved as **option (b): ERROR_CLASSES = NO** with rationale paragraph below the table. The envelope is informational signal; if the underlying problem ends the turn later, that turn-end failure gets whatever existing class applies. `classify_failure` stays untouched for this envelope. **`KNOWN_TASK_BLOCKED_REASONS` token still applies** — different surface (lead's `task_reassign` reason field).

### Sharpen 3 — Configuration surface subsection (7075c78)

New §5.6 names the four-fold surface (CLI flag / env var / `agents/<name>.json` key / capability-manifest field) for every v2 envelope tunable. Defaults to team-lead's suggested naming:
- `CLAUDE_ANYTEAM_WRAPPER_TOOL_FAILURE_WINDOW_S` (W=90s)
- `CLAUDE_ANYTEAM_APP_SERVER_IDLE_QUIET_WINDOW_S` (60s)
- `CLAUDE_ANYTEAM_TRANSPORT_ALIVE_CADENCE_S` (90s)
- `CLAUDE_ANYTEAM_WORKING_ON_CADENCE_S` (60s)
- `CLAUDE_ANYTEAM_SUBPROCESS_PRESSURE_DEBOUNCE_S` (60s)

Includes cross-reference to existing knobs and a lock-step requirement: if Phase A's impl PR changes any name, it must update §5.6 simultaneously.

### Net

Reviewer's three sharpens addressed in the RFC itself rather than deferred to Phase A. Approve status unchanged; merge whenever the owner is ready.

cc @opus-reviewer (no re-review needed unless you spot something).

PR #54, opened 2026-05-12
fix(#43): wrapper-side mitigation for codex sqlite WAL bloat

### LINKED ISSUE #43: claude-anyteam wrapper-side mitigation for codex CLI sqlite WAL bloat (#40 follow-up)

JonathanRosado:
## Summary

- add `claude_anyteam.codex_log_bloat` to stat Codex `logs_*.sqlite-wal`, apply the 100 MiB threshold, and run bounded `PRAGMA wal_checkpoint(TRUNCATE)` through sqlite locking APIs
- emit a pre-spawn typed `visibility_degraded` warning before every Codex App Server child start when WAL bloat is detected
- add `claude-anyteam diagnose --codex-log-bloat` plus docs/tests for the operator-facing diagnostic

## Investigation notes

- `codex --help` / `codex app-server --help` do not advertise a log-only redirect such as `CODEX_LOG_DIR`.
- Codex 0.128.0 does honor `CODEX_SQLITE_HOME`, but empirical launch with a temp home created both `logs_2.sqlite` and `state_5.sqlite` there. Because that is an all-sqlite-home redirect rather than a log-only redirect, this PR does not force it for teammates (preserves Codex harness state semantics). The diagnostic observes `CODEX_SQLITE_HOME` when the operator sets it.

## Proof of repro

Simulated the issue in a temp Codex sqlite home without touching the user's real `~/.codex`:

```text
$ truncate -s 200M "$tmp/logs_2.sqlite-wal"
$ uv run claude-anyteam diagnose --codex-log-bloat --codex-sqlite-home "$tmp"
claude-anyteam diagnose --codex-log-bloat mode=read-only
sqlite_home=/tmp/tmp.ePKENaDKWc source=explicit
threshold=100.0 MiB (104857600 bytes)

[codex-log-bloat]
status=degraded wal_files=1 bloated=1 total_wal=200.0 MiB max_wal=200.0 MiB
- /tmp/tmp.ePKENaDKWc/logs_2.sqlite-wal status=BLOATED wal=200.0 MiB db=-
```

Pre-fix sequence was the #40/#43 failure shape: no bloat-specific event before initialize progress/timeout:

```text
codex_app_server_mcp_config_prepared
app_server_initialize_progress ...
app_server_initialize_progress ...
turn_failed: JSON-RPC stdio process did not respond to initialize within 90.0s
```

## Proof of fix

With the same fake 200 MiB WAL pointed at `CODEX_SQLITE_HOME`, the pre-spawn hook fires before initialize:

```text
visibility_degraded severity=warn surface=codex_sqlite_wal_bloat max_wal_bytes=209715200 summary=Codex sqlite WAL bloat detected before App Server spawn; initialize may be delayed
app_server_initialize_completed severity=info surface=None max_wal_bytes=None summary=Codex App Server initialize completed in 0ms
turn_started severity=info surface=None max_wal_bytes=None summary=Codex App Server turn started
turn_completed severity=info surface=None max_wal_bytes=None summary=Codex App Server turn completed
```

The checkpoint path is covered with a real sqlite WAL fixture; it runs `PRAGMA wal_checkpoint(TRUNCATE)` and verifies the WAL is drained/deleted without direct file deletion.

## Tests

```text
uv run pytest -q tests/test_codex_log_bloat.py tests/test_app_server_mcp_config.py tests/test_app_server_initialize_visibility.py tests/test_diagnose_cli.py
25 passed in 1.50s

uv run pytest -q
1140 passed, 11 skipped, 2 deselected, 1 warning in 145.61s (0:02:25)
```

Fixes #43

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -162,6 +162,14 @@ The report combines `team-roster --json`, the rich manifest cache under `~/.clau
 
 `claude-anyteam diagnose` is read-only by default. The only mutating option is `--instrument-spawn`, which writes `env.CLAUDE_ANYTEAM_WRAPPER_MCP_DIAGNOSTICS=1` to `~/.claude/settings.json` so the next teammate spawn captures wrapper MCP diagnostics; use it only when the user explicitly asks to enable instrumentation.
 
+For Codex App Server initialize hangs that mention sqlite/WAL bloat (#43), run:
+
+```bash
+claude-anyteam diagnose --codex-log-bloat
+```
+
+This scans Codex's sqlite home (`CODEX_SQLITE_HOME`, else `CODEX_HOME`, else `~/.codex`) for `logs_*.sqlite-wal` files over the 100 MiB warning threshold. At runtime, claude-anyteam emits a typed `visibility_degraded` warning before spawning `codex app-server` when the threshold is exceeded and attempts a bounded `PRAGMA wal_checkpoint(TRUNCATE)` through sqlite's locking API; it never deletes Codex-owned log files directly.
+
 ## Shim configuration
 
 | Variable | Purpose |
--- a/src/claude_anyteam/app_server.py
+++ b/src/claude_anyteam/app_server.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Callable
 
 from . import logger
 from .jsonrpc_stdio import JsonRpcStdioClient, JsonRpcStdioError
@@ -19,9 +19,11 @@ class AppServerClient(JsonRpcStdioClient):
         codex_binary: str = "codex",
         extra_args: list[str] | None = None,
         env: dict[str, str] | None = None,
+        pre_start_hook: Callable[[], None] | None = None,
     ) -> None:
         self._codex_binary = codex_binary
         self._extra_args = list(extra_args or [])
+        self._pre_start_hook = pre_start_hook
         super().__init__(
             argv=[codex_binary, "app-server", *self._extra_args],
             env=env,
@@ -43,6 +45,13 @@ class AppServerClient(JsonRpcStdioClient):
         )
         self._error_cls = AppServerError
         self.last_thread_result: dict[str, Any] | None = None
+
+    def start(self) -> None:
+        """Start the child process after any wrapper-side preflight hook."""
+
+        if self._pre_start_hook is not None:
+            self._pre_start_hook()
+        super().start()
 
     # ---- transport health / restart ---------------------------------------
 
--- a/src/claude_anyteam/cli.py
+++ b/src/claude_anyteam/cli.py
@@ -170,7 +170,7 @@ def _build_run_parser() -> argparse.ArgumentParser:
             "  claude-anyteam team-roster  Print one-line-per-member team summary (flags ghosts and dead-pane members)\n"
             "  claude-anyteam team-config  Print resolved spawn-time config for a teammate (host model + adapter overrides)\n"
             "  claude-anyteam team-prune-dead  Remove members whose backing tmux pane is gone (use --yes to apply)\n"
-            "  claude-anyteam diagnose     Inspect substrate state: roster, manifests, visibility, wrapper MCP diagnostics\n"
+            "  claude-anyteam diagnose     Inspect substrate state; use --codex-log-bloat for Codex sqlite WAL checks\n"
             "  claude-anyteam status       One-screen team snapshot — roster, adapter overrides, incidents, last activity\n"
             "  claude-anyteam visibility-tail  Follow the live VisibilityEvent JSONL stream"
         ),
--- a/src/claude_anyteam/codex.py
+++ b/src/claude_anyteam/codex.py
@@ -49,6 +49,10 @@ from .env import (
     identity_env,
     env_first,
 )
+from .codex_log_bloat import (
+    checkpoint_bloated_codex_wals,
+    inspect_codex_log_bloat,
+)
 from .headless_visibility import HeadlessTurnVisibility, coerce_stream_text
 from .messages import VisibilityEvent
 from .prompts import TEAM_MESSAGING_BLOCK
@@ -1535,6 +1539,7 @@ def app_server_invoke(
     client = AppServerClient(
         codex_binary=codex_binary,
         env=identity_env(os.environ, team=settings_team, name=settings_agent),
+        pre_start_hook=lambda: _codex_log_bloat_pre_spawn(),
     )
 
     events: list[dict[str, Any]] = []
@@ -1611,6 +1616,84 @@ def app_server_invoke(
                     error=str(e),
                 )
         return event
+
+    def _codex_log_bloat_pre_spawn() -> None:
+        """Emit #43 pre-spawn visibility and run a bounded WAL checkpoint."""
+
+        try:
+            report = inspect_codex_log_bloat(env=os.environ)
+        except Exception as e:
+            logger.debug("codex.sqlite_wal_bloat.inspect_failed", error=str(e))
+            return
+        if not report.is_bloated:
+            return
+
+        report_payload = report.to_dict()
+        _emit_visibility_event(
+            kind="visibility_degraded",
+            severity="warn",
+            summary=(
+                "Codex sqlite WAL bloat detected before App Server spawn; "
+                "initialize may be delayed"
+            ),
+            visibility={
+                "mailbox": False,
+                "task_state": False,
+                "event_log": True,
+                "stderr": True,
+            },
+            payload={
+                "surface": "codex_sqlite_wal_bloat",
+                "action": "bounded_checkpoint_before_spawn",
+                "diagnose_command": "claude-anyteam diagnose --codex-log-bloat",
+                **report_payload,
+            },
+        )
+
+        try:
+            checkpoint_results = checkpoint_bloated_codex_wals(report, env=os.environ)
+        except Exception as e:
+            logger.debug("codex.sqlite_wal_bloat.checkpoint_failed", error=str(e))
+            return
+        if not checkpoint_results:
+            return
+        checkpoint_payload = {
+            "surface": "codex_sqlite_wal_checkpoint",
+            "results": [row.to_dict() for row in checkpoint_results],
+        }
+        all_ok = all(row.ok for row in checkpoint_results if row.attempted)
+        any_attempted = any(row.attempted for row in checkpoint_results)
+        if not any_attempted:
+            return
+        if all_ok:
+            _emit_visibility_event(
+                kind="turn_progress",
+                severity="info",
+                summary="Codex sqlite WAL checkpoint completed before spawn",
+                visibility={
+                    "mailbox": False,
+                    "task_state": False,
+                    "event_log": True,
+                    "stderr": True,
+                },
+                payload=checkpoint_payload,
+            )
+        else:
+            _emit_visibility_event(
+                kind="visibility_degraded",
+                severity="warn",
+                summary=(
+                    "Codex sqlite WAL checkpoint did not complete before "
+                    "App Server spawn"
+                ),
+                visibility={
+                    "mailbox": False,
+                    "task_state": False,
+                    "event_log": True,
+                    "stderr": True,
+                },
+                payload=checkpoint_payload,
+            )
 
     def _record(ev: dict[str, Any]) -> None:
         nonlocal tool_call_events
--- a/src/claude_anyteam/codex_log_bloat.py
+++ b/src/claude_anyteam/codex_log_bloat.py
@@ -1,0 +1,417 @@
+"""Detection and bounded mitigation for Codex CLI sqlite WAL bloat.
+
+Codex CLI stores its trace/state sqlite databases under the Codex sqlite home
+(defaulting to ``~/.codex``). Issue #43 tracks an upstream failure mode where
+``logs_*.sqlite-wal`` can grow large enough that ``codex app-server`` spends
+its JSON-RPC initialize budget inside sqlite WAL recovery/checkpoint work.
+
+This module keeps the claude-anyteam side small and harness-preserving:
+we inspect Codex-owned files, emit typed visibility when they look dangerous,
+and optionally ask sqlite to checkpoint through the public database API. We do
+not delete, rotate, or rewrite Codex logs directly.
+"""
+
+from __future__ import annotations
+
+import os
+import sqlite3
+import time
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Mapping
+
+CODEX_HOME_ENV = "CODEX_HOME"
+CODEX_SQLITE_HOME_ENV = "CODEX_SQLITE_HOME"
+
+CODEX_WAL_WARN_THRESHOLD_BYTES_ENV = (
+    "CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_WARN_THRESHOLD_BYTES"
+)
+CODEX_WAL_CHECKPOINT_ENV = "CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_CHECKPOINT"
+CODEX_WAL_CHECKPOINT_TIMEOUT_ENV = (
+    "CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_CHECKPOINT_TIMEOUT_S"
+)
+
+DEFAULT_CODEX_WAL_WARN_THRESHOLD_BYTES = 100 * 1024 * 1024
+DEFAULT_CODEX_WAL_CHECKPOINT_TIMEOUT_S = 10.0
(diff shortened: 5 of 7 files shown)


JonathanRosado (commented):
**Disposition: APPROVE.**

PR delivers on the lead's brief and the RFC's option-(a) framing ("land against existing taxonomy, refactor when `subprocess_pressure` lands as code"). Full pytest verified at **1140 passed / 11 skipped / 2 deselected** — matches architect's claim. Six new tests cover both proof-of-repro (sparse 200 MiB WAL detected as bloated) and proof-of-fix (envelope emitted AND positioned before `app_server_initialize_completed` in the event stream). The checkpoint path is exercised against a **real sqlite WAL fixture** (`test_checkpoint_path_truncates_real_sqlite_wal`) — actual `PRAGMA wal_checkpoint(TRUNCATE)` against a `journal_mode=WAL`, `autocheckpoint=0`, 256-row test DB, with assertion that the WAL file is drained without direct deletion.

Eight SHARPEN findings below — none blocking, all suitable for a follow-up tweak or for the eventual `subprocess_pressure` refactor PR. No BLOCK.

---

### Verified passes against charter

- **Invariant 2 / 7 (typed events, taxonomy):** Bloat envelope uses `kind="visibility_degraded"` with `payload.surface="codex_sqlite_wal_bloat"` — option (a) per RFC's guidance, riding on existing taxonomy. Forward refactor to `kind="subprocess_pressure"` is a one-cell change in `_codex_log_bloat_pre_spawn` when that envelope lands. No four-leg registration needed for this PR (the envelope is `visibility_degraded` which is already registered).
- **Invariant 8 (fan-out symmetry):** event_log + stderr fan-out is symmetric. Mailbox is intentionally `False` — see S1 below for the question.
- **Invariant 9 (PRAGMA safety):** Checkpoint uses sqlite's own locking API. `sqlite3.connect(..., timeout=timeout_s)` + `PRAGMA busy_timeout` covers concurrent-writer races (sqlite serializes properly; we get `busy=1` if codex holds the writer lock and we run out of budget). `set_progress_handler` with a `time.monotonic() >= deadline` check provides a hard wall-clock bound mid-checkpoint via `OperationalError("interrupted")`. Three failure modes (`busy`, `timeout`, `error`) each map to distinct status codes in `CodexWalCheckpointResult`. Per-call default 10s is bounded.
- **Invariant 11 (timer skepticism):** The new wall-clock timer is bounded-I/O class (sqlite checkpoint budget), not stall-detection. Per RFC §1.2 / §0 framing, that's the right side of the line. The pre-spawn hook itself is a state-change trigger (every spawn = new check), not a periodic timer.
- **Invariant 12 (Mode A/B):** Not directly applicable here (this is pre-spawn, not mid-turn), but the design pattern is consistent: typed envelope fires only on the actual state change (`if not report.is_bloated: return`), not on a wall-clock window.
- **Proof-of-fix shape:** `test_pre_spawn_warning_emits_visibility_degraded_before_initialize` asserts the envelope IS emitted AND its event-sequence position is before `app_server_initialize_completed`. Verifies behavior change, not absence-of-crash. Good test pattern; matches PR #42's precedent.

---

### Sharpen findings (8 — none blocking)

**S1. `visibility={"mailbox": False, ...}` on the bloat envelope — asymmetric with PR #42 precedent.** PR #42's `emit_initialize_timeout_visibility_degraded` sets `mailbox=True` (`protocol_io.py:949-953`) because the timeout BLOCKS the turn and the lead must see it. PR #54's bloat envelope sets `mailbox=False` because the bloat is recoverable via the in-line checkpoint. That's a defensible choice (don't crowd the inbox with recoverable conditions) but it's IMPLICIT. Two questions for the architect to answer in a code comment or follow-up:
   - If the checkpoint fails (`status="busy"` or `status="timeout"`), the SECOND `visibility_degraded` envelope still has `mailbox=False`. At that point the bloat will likely cause an initialize timeout — should THIS envelope be `mailbox=True` so the lead sees it before the eventual timeout fires? Or is the eventual `app_server_initialize_timeout` (already `mailbox=True`) the lead-facing surface?
   - When `mailbox=False`, the lead's only visibility is via `visibility_tail` or `claude-anyteam diagnose --codex-log-bloat`. The skills/help text in the Phase C migration (RFC §8 Phase C) should teach the lead to check these surfaces — currently nothing mentions the bloat envelope is invisible to the inbox flow.

   Suggest: add a code comment on the `visibility` dict justifying `mailbox=False` for the WARN, AND consider `mailbox=True` for the checkpoint-FAILED branch (where the bloat will likely time out anyway).

**S2. `kind="turn_progress"` for the checkpoint-success envelope is a semantic mismatch.** Looking at `_codex_log_bloat_pre_spawn`:

   ```python
   if all_ok:
       _emit_visibility_event(
           kind="turn_progress",
           severity="info",
           summary="Codex sqlite WAL checkpoint completed before spawn",
           ...
       )
   ```

   `turn_progress` is documented as "model progress during a turn" (`messages.py:365` Literal + adapter loop semantics). Pre-spawn checkpoint completion is neither in-turn nor model-emitted. PR #42's precedent for pre-turn-start signals is `app_server_initialize_progress` / `app_server_initialize_completed`. Using `turn_progress` here will surface as "model is making progress" in any consumer filtering by event kind — false positive.

   Options:
   - Use `kind="visibility_degraded"` with `severity="info"` and a different summary indicating recovery (mirrors the WARN envelope structure).
   - Use a new dedicated kind (e.g., `codex_sqlite_wal_checkpoint`) but that triggers the four-leg taxonomy work the PR is explicitly deferring.
   - Reuse `app_server_initialize_progress` if the semantics fit (probably not — checkpoint is before initialize).

   Cleanest: `kind="visibility_degraded"`, `severity="info"`, same `payload.surface="codex_sqlite_wal_checkpoint"` already specified.

**S3. Env vars not registered in `env.py`.** The three new env var names (`CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_WARN_THRESHOLD_BYTES`, `CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_CHECKPOINT`, `CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_CHECKPOINT_TIMEOUT_S`) live as constants inside `codex_log_bloat.py` rather than in `env.py`. PR #42 set the precedent (`APP_SERVER_INITIALIZE_TIMEOUT_ENV`, etc. live in `env.py`) so all env var names are discoverable from one location. Inconsistency. Move them to `env.py`.

**S4. No upper-bound range validation on env values (invariant 4).** `_int_env` returns `max(0, value)`; `_float_env` returns `max(0.001, value)`. A user setting `CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_WARN_THRESHOLD_BYTES=0` causes every codex spawn to fire the bloat envelope (false-positive engine). PR #52 has explicit `if not (60.0 <= turn_timeout_s <= 3600.0): raise ValueError(...)` style range validation. PR #54 doesn't. Recommend:
   - Threshold: lower bound 1 MiB (no point warning on smaller WALs), upper bound something like 10 GiB (clearly an operator error if higher).
   - Checkpoint timeout: lower bound 1s, upper bound 60s.

**S5. Per-WAL timeout, not aggregate-bounded.** `checkpoint_bloated_codex_wals` iterates each bloated WAL with the same `timeout_s` budget per WAL. With N=5 bloated WALs, total pre-spawn delay is N × 10s = 50s, stacked on top of the eventual `app_server_initialize_timeout` (90s). In practice codex creates one log DB (`logs_2.sqlite`), so the multi-WAL case is mostly theoretical — but the code allows it. Add an aggregate deadline shared across all WALs so total pre-spawn delay is bounded. Suggest `total_budget = timeout_s` (same as per-WAL today, but applied as a shared deadline across all checkpoints).

**S6. Test coverage gap on `checkpoint_codex_wal` failure branches.** The `busy`, `timeout`, `error`, `missing_db` branches all have distinct code paths but no test exercises them. Suggest adding:
   - `test_checkpoint_returns_missing_db_when_database_absent` — pass a WAL whose `database_path` doesn't exist.
   - `test_checkpoint_returns_busy_when_writer_lock_held` — open a long-running transaction in a fixture, run checkpoint, expect `status="busy"`.
   - `test_checkpoint_returns_timeout_when_progress_handler_fires` — set `timeout_s=0.001`, verify status="timeout".

   The success path test is already there; these three would round out the matrix.

**S7. Forward-compat with future `subprocess_pressure` envelope payload (lead's specific concern).** The current payload uses `surface=codex_sqlite_wal_bloat` and `diagnose_command`. RFC §5.3's future `subprocess_pressure` envelope wants `pressure_source: "sqlite_wal_bloat"` (per the lead's mention) and `remediation` (per RFC §5.3 payload spec). When the refactor lands, consumers reading `payload.surface` will need migration. Suggest adding aliases NOW so the future top-level-kind change is just a `kind` value change without payload field rename:

   ```python
   payload={
       "surface": "codex_sqlite_wal_bloat",
       "pressure_source": "sqlite_wal_bloat",  # alias for future subprocess_pressure
       "action": "bounded_checkpoint_before_spawn",
       "diagnose_command": "claude-anyteam diagnose --codex-log-bloat",
       "remediation": "claude-anyteam diagnose --codex-log-bloat",  # alias for future
       "hint": f"codex sqlite WAL is {report.total_wal_bytes // (1024*1024)}MB; startup may be slow",
       **report_payload,
   },
   ```

   Then when `subprocess_pressure` ships, the change is exactly one line (`kind="visibility_degraded"` → `kind="subprocess_pressure"`) — `surface`/`diagnose_command` can be removed in the same PR but consumers reading the new aliases work in both versions.

**S8. No CHANGELOG entry.** PR #52 added an "## [Unreleased]" section. PR #54 doesn't extend it. Should note the new `--codex-log-bloat` diagnose subcommand and the three new env vars under "### Added".

---

### Side note (PR #52 BLOCK status)

Noticed two new commits on `feat/bump-default-timeouts` since my last PR #52 review: `38a3287 fix(#52 block): decouple non_progress_interrupt_s from warn flag` and `3738a07 test(#52): negative + positive coverage for warn=None decoupling`. The architect chose option (b) — decouple — rather than option (a) CHANGELOG note for my BLOCK finding. That's a stronger fix. Not reviewing those commits here (separate PR/disposition), but flagging so the lead can pull the PR #52 review forward when ready.

---

### Net

The architectural shape is right. The implementation is harness-preserving (no log deletion, uses sqlite's own API, observes operator env vars rather than forcing them). The visibility hook fires at the correct point in the spawn sequence and the proof-of-fix test asserts it. Eight sharpens are appropriate for follow-up; none of them block the merge.

— `opus-reviewer`

PR #55, opened 2026-05-12
fix(#48): refuse bare routed prefixes (codex-/gemini-/kimi-) without per-teammate config

### LINKED ISSUE #48: Agent tool + skill don't prevent the silent failure: `codex-*` name without `team-agent` config spawns native Claude

JonathanRosado:
Fixes #48.

## Repro / before
A routed teammate name such as `codex-ghost` / `gemini-ghost` / `kimi-ghost` could be used in `Agent(...)` before a matching `~/.claude/teams/<team>/agents/<name>.json` existed. The spawn shim treated the missing per-teammate file as a no-op and continued, which made a bare-prefix spawn look successful even when the lead skipped the documented `claude-anyteam team-agent ...` setup. The regression test now drives that exact missing-config path.

## Fix / after
The spawn shim now soft-refuses routed prefixes when the per-teammate config file is absent, emitting a structured `spawn_shim.bare_prefix_refused` JSON error with the expected config path, a copy-pasteable `claude-anyteam team-agent <name> --team <team> --model <model> --effort <effort>` command, and the explicit escape hatch `CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX=1`. A companion test proves the override still permits adapter-default dispatch when deliberately enabled.

## Files changed
- `src/claude_anyteam/spawn_shim.py`
- `src/claude_anyteam/env.py`
- `tests/test_spawn_shim.py`
- `hooks/session-start.sh`
- `tests/test_plugin_bundle.py`
- `.claude-plugin/plugin.json`
- `.claude-plugin/marketplace.json`
- `skills/help/SKILL.md`
- `docs/configuration.md`

## Proof
- `uv run pytest -q tests/test_spawn_shim.py tests/test_plugin_bundle.py` → 41 passed


JonathanRosado (commented):
**Disposition: APPROVE.**

PR delivers on the #48 brief: replaces the silent native-Claude fallback for missing-config bare routed prefixes with a structured refusal. Two real SHARPEN findings (S1 override audit gap, S2 missing CHANGELOG) and three minor sharpens. No BLOCK.

### Verified clean

- **Full pytest at `e959546`: 1135 passed / 11 skipped / 2 deselected.** +1 net from 1134 baseline, consistent with two new spawn-shim tests minus one removed (`test_agent_config_missing_file_is_noop` deleted because the noop branch no longer exists). Targeted `test_spawn_shim.py + test_plugin_bundle.py` runs at 41 passing — matches architect's claim.
- **Two new tests cover both directions of the regression surface:**
  - `test_routed_prefix_without_agent_config_soft_refuses` — proof-of-fix: exit code 2, NO execv (calls == []), structured stderr event with `event=spawn_shim.bare_prefix_refused`, `route=codex`, `error_class=missing_agent_config`, `config_path`, `suggested_command`, `override_env` fields. The full payload shape matches the `spawn_shim.bare_prefix_refused` schema the lead asked about.
  - `test_bare_prefix_override_allows_dispatch_without_config` — proof of override: exit code 0, execv proceeds for codex route, dispatch event emitted.
- **Cross-branch contamination resolved.** `da2cf40` and `e959546` are content-identical (same files, same line counts, same author/timestamp). The architect already pushed `b5d80f3 revert: da2cf40 (carried by cross-branch contamination; ships in PR #55)` to `feat/bump-default-timeouts`. PR #55 is the canonical home; I reviewed the content here only.

### Charter passes

- **Invariant 3 (spawn_shim native-fallthrough signal):** Replaces silent fallback with a structured stderr-JSON event AND exit code 2. The event carries `error_class`, `error_detail`, `config_path`, `suggested_command`, `override_env`, `override_hint`, `issue` fields — fully diagnostic, lead-actionable.
- **Invariant 4 (no ranges to validate here):** The new env var `CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX` is a boolean (truthy parsing via `_env_flag_enabled`); no numeric range to validate. ✓
- **Invariant 7 (typed taxonomy)** — partial, but defensibly so: the event is stderr-JSON (`json.dumps(record)`), NOT a typed `VisibilityEvent`. The spawn_shim is **stdlib-only by design** (line 3 comment: "This module is intentionally stdlib-only so it can be imported standalone") — it runs before the wrapper venv is set up and routes to native Claude when no adapter is available, so it can't depend on `protocol_io.VisibilityEvent`. The new event is consistent with the existing `spawn_shim.dispatch` stderr surface. A future "spawn-shim event projector" tool could lift these stderr events into typed envelopes; out of scope for this PR.
- **Override safety:** `_env_flag_enabled` accepts `{"1", "true", "yes", "on"}` — case-insensitive, whitespace-stripped. Reasonable parsing.

### Findings

**S1. (SHARPEN — invariant 3 echo) Override is unaudited.** When `CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX=1` is set + no config file exists, `_maybe_refuse_bare_routed_prefix` returns `None` silently and the spawn proceeds. The existing `spawn_shim.dispatch` event IS emitted — but it's INDISTINGUISHABLE from a normal config-backed spawn (the only signal is the absence of an `agent_config` field, which is identical to a spawn where the config exists but is empty). A lead auditing spawn events can't tell:

  - "User has full per-teammate config" vs
  - "User has no per-teammate config but is bypassing via override"

without correlating dispatch records with filesystem checks. **This is the same anti-pattern shape the PR is fixing** — invisible fallback to defaults. Suggested fix:

  - Add a separate `spawn_shim.bare_prefix_allowed_via_override` event when override is enabled AND config is missing, OR
  - Add an `override_used: true` (and `override_env: "CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX"`) field to the dispatch event when the override branch is taken.

  Option (a) is cleaner — same event taxonomy as the refuse event, just for the override-engaged case. One ~10-line addition.

**S2. (SHARPEN — invariant 9) No CHANGELOG entry for a backwards-incompatible behavior change.** Before this PR, a routed-prefix spawn without per-teammate config silently proceeded (the #48 bug). After this PR, the same spawn returns exit code 2 unless `CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX=1` is set. Users relying on the (broken) silent fallback will break on upgrade. The "## [Unreleased]" section (added by PR #52) should be extended with:

  > **Routed-prefix teammate spawns (`codex-*`, `gemini-*`, `kimi-*`) now require a per-teammate config file** at `~/.claude/teams/<team>/agents/<name>.json` created via `claude-anyteam team-agent`. Spawning a bare routed prefix without that file returns exit code 2 with a structured `spawn_shim.bare_prefix_refused` stderr event (issue #48). Restore the prior behavior with `CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX=1` (and see S1 — currently this override is unaudited, will improve in a follow-up).

  Two paragraphs in the CHANGELOG would close the invariant-9 gap.

**S3. (MINOR) Deleted test without a replacement comment.** `test_agent_config_missing_file_is_noop` is removed because the noop case no longer exists. That's correct — but a future reader scanning git blame on the test file will wonder where the coverage went. Suggest adding a one-line docstring on `test_routed_prefix_without_agent_config_soft_refuses` like *"Replaces the removed test_agent_config_missing_file_is_noop — the noop branch is gone; routed-prefix missing config is now a hard refusal (#48)."*

**S4. (MINOR — out of scope here, file as follow-up?) Native-prefix typo footgun is adjacent but distinct.** The PR correctly excludes `_claude_route` from the refuse check (native dispatch is always pass-through). But a typoed name like `cladue-foo` (no recognized prefix) silently falls through to native claude — same `_log_dispatch("native", parsed.agent_name, binary)` path that #48 was concerned with. Different cause (typo vs missing config) but similar invisibility. Worth filing as #56 if it isn't already on the radar; not blocking #55 since #48's brief was specifically about the prefix-matched-no-config case.

**S5. (COSMETIC) Proof-of-repro in PR body is prose, not literal.** PR #54 showed actual before/after command-line outputs. PR #55's "Repro / before" is descriptive. The regression test is concrete proof-of-fix (verified in my run), but the PR body would be stronger with the actual stderr JSON the new code produces alongside the prose. Minor.

### Net

The architectural shape is right (stdlib-only spawn_shim emits structured stderr; exit code 2 means caller sees the failure). Tests cover both directions. Documentation surfaces (help skill, session-start hook, configuration.md) are tightened consistently. The two real sharpens (S1 unaudited override, S2 CHANGELOG) are cheap to address — recommend the architect land them in a fixup commit before merge, but neither blocks.

— `opus-reviewer`

PR #56, opened 2026-05-12
fix(#40): phase 2 codex app-server cold-start mitigation

JonathanRosado:
## Summary

- add a bounded cross-process Codex App Server cold-start gate: randomized pre-start jitter, file-lock serialization, and short post-start cooldown
- retry JSON-RPC `initialize` timeouts once with typed `turn_warning(surface=app_server_initialize_retry)` before surfacing the existing failure path
- document the new knobs and cover the gate + retry paths with unit tests

## Proof of repro / baseline

The original #40 failing sequence from the issue was:

```text
codex_app_server_mcp_config_prepared
... no codex_app_server_session_start ...
turn_failed: JSON-RPC stdio process did not respond to initialize within 600.0s
```

On this host after isolating sqlite state (so #43 WAL bloat is not part of the measurement), I ran a wrapper-level equivalent of a two-`codex-*` Agent batch: two concurrent `AppServerClient` cold starts + `initialize`, N=5 pairs, with a temp `CODEX_SQLITE_HOME` to avoid touching the user's real `~/.codex` logs.

Pre-fix baseline did not reproduce the full hang on this now-clean host, but established timing:

```text
run 1: client1 ok=True elapsed=0.44s | client2 ok=True elapsed=0.37s
run 2: client1 ok=True elapsed=0.42s | client2 ok=True elapsed=0.48s
run 3: client1 ok=True elapsed=0.42s | client2 ok=True elapsed=0.47s
run 4: client1 ok=True elapsed=0.97s | client2 ok=True elapsed=1.07s
run 5: client1 ok=True elapsed=1.19s | client2 ok=True elapsed=1.17s
```

A unit repro simulates the actionable Phase 2 failure mode: first initialize raises `JSON-RPC stdio process did not respond to initialize...`; before this PR, `app_server_invoke` returned the failure immediately and loop-side code marked the task/prose turn failed.

## Proof of fix

Post-fix concurrent cold-start timings with the gate enabled, same N=5 pairs and temp sqlite home:

```text
run 1: client1 ok=True elapsed=0.68s | client2 ok=True elapsed=0.49s
run 2: client1 ok=True elapsed=0.62s | client2 ok=True elapsed=0.44s
run 3: client1 ok=True elapsed=0.55s | client2 ok=True elapsed=0.73s
run 4: client1 ok=True elapsed=0.78s | client2 ok=True elapsed=0.65s
run 5: client1 ok=True elapsed=0.68s | client2 ok=True elapsed=0.86s
[...]

PR #60, opened 2026-05-12
feat(#49): emit unrecovered wrapper-tool failure signal

JonathanRosado:
## Summary
- add top-level `wrapper_tool_failure_unrecovered` visibility kind for #49
- implement Codex App Server Mode A/B discriminator: after a failed wrapper MCP tool event, emit only if no later `turn_progress`/`tool_event`/`artifact_event`/agentMessage delta appears within W
- expose W as `--wrapper-tool-failure-window-s`, `CLAUDE_ANYTEAM_WRAPPER_TOOL_FAILURE_WINDOW_S`, and per-teammate `wrapper_tool_failure_window_s`
- register the lead-action `KNOWN_TASK_BLOCKED_REASONS` token while deliberately leaving `diagnostics.ERROR_CLASSES` untouched
- document the new configuration surface

## Tests
- `uv run pytest -q` — 1142 passed, 11 skipped, 2 deselected
- targeted pre-full-suite checks also passed for config/parser/taxonomy/spawn-shim/team-cli and the Mode A/B discriminator tests

Refs #49

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -19,6 +19,10 @@ claude-anyteam \
   --effort <level>          # low | medium | high | xhigh (default: Codex's default)
   --plan-mode               # opt into plan approval mode
   --no-app-server           # opt out of App Server mode (use fresh-exec instead)
+  --turn-timeout-s <sec>    # Codex App Server turn cap (default: 900; range [60, 3600])
+  --non-progress-warn-s <sec>             # soft progress watchdog (default: 300; range [60, 900])
+  --non-progress-interrupt-s <sec>        # opt-in hard progress interrupt (default: unset)
+  --wrapper-tool-failure-window-s <sec>   # #49 unrecovered wrapper-tool signal window (default: 90; range [60, 300])
   --poll-s <float>          # inbox poll interval in seconds (default: 1.5)
   --color <name>            # display color in peer DMs (default: cyan)
   --log <level>             # debug | info | warn | error (default: info)
@@ -80,6 +84,10 @@ Every flag has an equivalent env var:
 | `CLAUDE_ANYTEAM_APP_SERVER_START_GATE_COOLDOWN_S` | short post-start serialization window before another Codex App Server spawn proceeds (default `0.25`) |
 | `CLAUDE_ANYTEAM_APP_SERVER_INITIALIZE_RETRIES` | retry count for Codex App Server initialize timeouts before surfacing failure (default `1`) |
 | `CLAUDE_ANYTEAM_APP_SERVER_INITIALIZE_RETRY_BACKOFF_S` | backoff before the initialize retry (default `2.0`) |
+| `CLAUDE_ANYTEAM_TURN_TIMEOUT_S` | `--turn-timeout-s` |
+| `CLAUDE_ANYTEAM_NON_PROGRESS_WARN_S` | `--non-progress-warn-s` |
+| `CLAUDE_ANYTEAM_NON_PROGRESS_INTERRUPT_S` | `--non-progress-interrupt-s` |
+| `CLAUDE_ANYTEAM_WRAPPER_TOOL_FAILURE_WINDOW_S` | `--wrapper-tool-failure-window-s` |
 | `CLAUDE_ANYTEAM_POLL_S` | `--poll-s` |
 | `CLAUDE_ANYTEAM_COLOR` | `--color` |
 | `CLAUDE_ANYTEAM_LOG` | `--log` |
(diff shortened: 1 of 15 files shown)


JonathanRosado (commented):
**Disposition: APPROVE.**

This is the **strongest implementation PR in the recent batch** — the implementer internalized the converging-pattern feedback the team-lead routed across #54/#55/#56 and applied it correctly. Five sharpens flagged below, all non-blocking. The §5.1 verbatim-fidelity is honest (the four-kind disjunction collapses to three observable event kinds because agentMessage normalizes to `turn_progress` — verified independently against `_visibility_for_app_server_item`'s agentMessage handler).

### Verified clean

- **Full pytest 1142 passed / 11 skipped / 2 deselected** (+6 from 1136 baseline). Targeted: both `wrapper_tool_failure_*` tests pass cleanly.
- **Four-leg taxonomy properly implemented (invariant 2):**
  - `VisibilityEventKind` Literal — `"wrapper_tool_failure_unrecovered"` added at `messages.py:401` with justification comment ✓
  - `parse_protocol_text` dispatch — added at `messages.py:516` ✓
  - `KNOWN_TASK_BLOCKED_REASONS` — token added with comment justifying it as a lead-action token, NOT ERROR_CLASSES ✓
  - `ERROR_CLASSES` — NOT added, per RFC §5.1 sharpen #2 option (b). The implementer engaged with my RFC review explicitly: `_wrapper_tool_error_class` is forensic-payload-only, never feeds `classify_failure`. ✓
- **Configuration-surface lock-step (invariant 11 / RFC §5.6) — 4 of 5 legs:**
  - `env.py` — `WRAPPER_TOOL_FAILURE_WINDOW_ENV` registered as central constant ✓ (**addresses S6 from PRs #54/#56**)
  - `config.py` — explicit bilateral range validation `if not (60.0 <= v <= 300.0): raise ValueError` ✓ (**addresses S5 from PRs #54/#56** — both bounds enforced, matches PR #52's pattern)
  - `cli.py` — `--wrapper-tool-failure-window-s` flag ✓
  - `team_cli.py` — `--wrapper-tool-failure-window-s` flag + included in the "at least one of" required-flag check ✓
  - `spawn_shim.py` — `_AGENT_CONFIG_KEYS` includes the new key for agents/<name>.json forwarding ✓
  - `loop.py` — settings forwarded to both `_invoke_codex_prose` and `_execute_task_app_server` ✓
  - `capabilities.py` — **NOT extended** (see S3 below)
- **Envelope visibility flags `mailbox=True, task_state=True, event_log=True, stderr=True`** ✓ — **addresses S3 from PRs #54/#56** (the recurring `mailbox=False` pattern). The implementer correctly identified this as a lead-actionable signal.
- **§0 separation:** the emission path is signal-only. No `client.turn_steer`, no `client.turn_interrupt`, no kill. The `payload.recovery_hint_dispatched=False` field is informational; the wrapper doesn't dispatch hints. ✓

### Verified §5.1 fidelity

The RFC pseudocode says recovery is signaled by any of `(turn_progress | tool_event | artifact_event | agentMessage byte delta)`. The implementation uses `_WRAPPER_TOOL_RECOVERY_EVENT_KINDS = frozenset({"turn_progress", "tool_event", "artifact_event"})`.

Independently verified the equivalence: at `codex.py:_visibility_for_app_server_item`, when an `agentMessage` item arrives, the wrapper emits `make_event(kind="turn_progress", payload={"agent_message_bytes": len(text), ...})`. So agentMessage byte deltas DO surface as `turn_progress` envelopes — the four-kind disjunction collapses cleanly into three observable envelope kinds. **The implementation is semantically equivalent to the RFC pseudocode.**

The recovery detection runs in `_emit_visibility_event` via sequence-number comparison: `event.seq > pending.failed_event_seq` means "emitted after the failure," equivalent to RFC's temporal "since failure" check. ✓

### Sharpens (5 — none blocking)

**S1. Boundary tests missing (lead's specific ask).** The two new tests cover Mode B at gap=W+1 (91s) and Mode A with recovery at 30s. Missing:

- **Gap = W-1 (recovery just before timeout):** Failure at t=0, recovery event at t=89, no further events. Discriminator should NOT fire. Easy to add.
- **Multiple failures within W:** Failure A at t=0, failure B at t=30 (B is a `tool_event` with `severity=error`), then success at t=60, no further events. Should produce ZERO `wrapper_tool_failure_unrecovered` envelopes (both A and B recovered by later events).

  **Subtle case worth covering**: failure A at t=0, failure B at t=30. Per the recovery-detection logic (`_emit_visibility_event` checks ANY tool_event kind including error-severity), failure B's emission marks failure A as recovered. **This is per RFC §5.1.1's pseudocode** ("a retry counts as progress, success or failure") but it's worth a test to pin the semantic. A "failure-retry-loop" sequence (A→B→C→D→silence) should produce exactly ONE envelope for D (the final failure that wasn't recovered).

**S2. Recovery predicate is a parallel implementation, not a shared helper.** `_WRAPPER_TOOL_RECOVERY_EVENT_KINDS = frozenset({"turn_progress", "tool_event", "artifact_event"})` is the implementer's reimplementation of the RFC-cited `codex.py:2196-2199` "observable progress" predicate. The original predicate tracks `last_message_len`, `last_tool_count`, and notification delta directly; the new predicate matches on envelope kinds. They're semantically equivalent today, but they're two source-of-truth points that can drift if a future PR adds a fifth progress signal (e.g., a new envelope kind for codex-app-server-internal "thinking" events).

Suggested: lift the recovery-kind set into a module-level constant or shared helper that both predicates consume. Doesn't have to happen here — could be a follow-up — but worth flagging so it doesn't quietly drift.

**S3. Capability-manifest entry not added (lead's "4th of 4 legs").** The team-lead's review brief explicitly listed "capability-manifest field" as part of the four-fold config surface. The implementer added the env var, CLI flag, agents.json key, and config range validation (a richer 5-of-5 from the user-facing perspective) but didn't extend `capabilities.py`'s `_BASE_CAPABILITY_MANIFEST` to declare the new tunable.

Looking at the existing pattern: `soft_non_progress_watchdog` declares `non_progress_warn_s` + `non_progress_interrupt_s` as properties so peers can introspect the configuration. The new window is analogous and should arguably get a `wrapper_tool_failure_discriminator` (or similar) capability entry with the `wrapper_tool_failure_window_s` property + range. Especially since the RFC §5.1 envelope is meant to be cross-peer-actionable per north-star §3.

Reasonable to defer to a follow-up if you'd rather not gate on this — but it IS the leg the lead's brief called out.

**S4. Multiple-failures-for-same-tool emit multiple envelopes.** Per the design: if `mcp_anyteam_read_file` fails repeatedly (e.g., the model keeps trying a file path that doesn't exist, in a retry loop where each retry is followed by another retry rather than a non-failure event), and the retries are W seconds apart, EACH retry produces its own `wrapper_tool_failure_unrecovered` envelope. The lead gets N envelopes for what's semantically one stuck loop.

Not a bug per the RFC, but worth a payload field like `prior_unrecovered_count` so the lead can recognize the pattern from envelope-count alone instead of correlating tool_name across envelopes. Or debounce by tool_name within a longer window. Design choice — flag for the implementer/architect's preference.

**S5. CHANGELOG entry still missing.** Fifth PR in the recent batch (#54, #55, #56, #60 — and PR #52 v2 had its sharpen handled separately) without extending "## [Unreleased]". This is the pattern the team-lead's "consolidating PR" suggestion is meant to clean up — leaving this as a flag for whoever picks up the consolidator.

### Observation: implementer internalized prior feedback

The pattern is clean to see across this PR:
- **S6 (env.py)** from PRs #54/#56 → ADDRESSED: `WRAPPER_TOOL_FAILURE_WINDOW_ENV` lives in `env.py`.
- **S5 (range validation)** from PRs #54/#56 → ADDRESSED: bilateral `[60, 300]` enforced in `config.py:from_env`.
- **S3 (mailbox=False)** from PRs #54/#56 → ADDRESSED: `mailbox=True` here.
- **S2 (semantic-kind mismatch)** from PRs #54/#56 → ADDRESSED: new top-level kind, not riding on `visibility_degraded` or `turn_progress`.
- **RFC §5.1 sharpen #2 (ERROR_CLASSES question)** → ADDRESSED: implementer explicitly chose NOT to add an ERROR_CLASSES entry, with comment justifying.

The remaining gaps (S1 boundary tests, S3 capability-manifest leg, S5 CHANGELOG) are smaller and unrelated to the prior pattern. The implementer demonstrably reads review feedback and applies it.

### Net

PR is clean to merge. The four-leg taxonomy is done correctly, the configuration-surface lock-step is largely done, the §5.1 RFC semantics match the wrapper's existing progress-signal taxonomy, and the envelope flags are right. Five sharpens are all suitable for a tiny follow-up or for the consolidating PR the lead mentioned will land across #54/#55/#56/#60.

— `opus-reviewer`

Issue #58, opened 2026-05-12
tooling friction: gh pr edit --body silent-no-op due to projects-classic GraphQL sunset


JonathanRosado:
## Symptom

`gh pr edit <num> --body[-file]` no-ops silently when the PR has any (even empty) projects-classic linkage in its GraphQL graph. The CLI exits with a deprecation **warning**:

```
GraphQL: Projects (classic) is being deprecated in favor of the new Projects experience, see: https://github.blog/changelog/2024-05-23-sunset-notice-projects-classic/. (repository.pullRequest.projectCards)
```

Despite reading like a warning, the underlying mutation is rejected and the body is **not** updated. `gh pr view <num> --json body` confirms the PR body unchanged.

## Observed 2026-05-12

While shipping the PR #52 contamination cleanup (per team-lead's authorized option-a path), I needed to put a contamination + revert note at the top of PR #52's description. Two attempts:

```bash
gh pr edit 52 --body "...long body..."
gh pr edit 52 --body-file /tmp/pr52-body.md
```

Both printed only the projects-classic deprecation message. PR body did not change. `gh pr view 52 --json body` confirmed.

## Workaround used

Posted a pinned **top-of-PR comment** instead. Less ideal because:
- Comments scroll out of the "important context" header position over time.
- Reviewers who land on the PR via deep-link to a commit may miss the comment entirely.
- The PR description is canonical; comments are conversational.

Comment URL for record: https://github.com/JonathanRosado/claude-anyteam/pull/52#issuecomment-4433108335

## Class of problem

This is tooling friction in the team workflow — anyone trying to programmatically update PR bodies for any reason (CHANGELOG cross-references, contamination notes, generated content) hits this. Plausible affected automations: `auto-release.yml`, any future PR-description-templating logic, this repo's `gh pr create` heredoc bodies (which work since `pr create` is a different mutation path).

## Suggested next steps

Likely needs one of:

1. **gh CLI upgrade** to a version past the projects-classic sunset. Repo currently using whatever's installed on `/home/rosad/...`; `gh --version` output would confirm.
2. **GitHub API behavior change** — if newer `gh` still hits this, the underlying GraphQL mutation needs updating to drop the `projectCards` selection from its PR-edit query.
3. **In the meantime**, document the workaround (pinned top comment) in any team-process docs that touch PR-description updates.

[...]

Issue #63, opened 2026-05-12
spawn_shim: typoed routed-prefix names (cladue-, geimini-, etc.) silently fall to native Claude


JonathanRosado:
Follow-up from opus-reviewer PR #55 S4.

Failure mode: typoed routed-prefix teammate names like `cladue-foo`, `geimini-foo`, or similar miss the spawn shim's routed-prefix regexes (`claude-`, `codex-`, `gemini-`, `kimi-`). Because they do not match any routed prefix, the shim treats them as ordinary native-Claude teammate names and silently falls through to native Claude.

This is adjacent to #48's bare routed-prefix footgun, but it has a different cause: #48 covered correctly-spelled routed prefixes with missing per-teammate config; this issue covers incorrectly-spelled routed-prefix-like names that never enter the routed-prefix guard at all.

Potential fix direction: detect close/typoed routed-prefix names before native fallback and refuse with a structured diagnostic that points users to the supported prefixes and `claude-anyteam team-agent` setup flow.

PR #64, opened 2026-05-12
fix(#48 + #49 follow-up): override-audit event + boundary tests + capability manifest + CHANGELOG

JonathanRosado:
Body-as-comment workaround for #58: see the first PR comment for the full summary, tests, and checklist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,13 @@ All notable changes to claude-anyteam are documented here. Format follows [Keep 
 
 ## [Unreleased]
 
+### Added
+
+- **`wrapper_tool_failure_unrecovered` visibility envelope** (issue #49 / PR #60) surfaces Codex App Server wrapper-MCP tool failures that are not followed by `turn_progress`, `tool_event`, or `artifact_event` recovery activity within the discriminator window. The window is configurable with `--wrapper-tool-failure-window-s`, `CLAUDE_ANYTEAM_WRAPPER_TOOL_FAILURE_WINDOW_S`, and the per-teammate agents-file key `wrapper_tool_failure_window_s`.
+
 ### Changed
 
+- **Breaking #48 routed-prefix spawn guard:** spawning a `codex-*`, `gemini-*`, or `kimi-*` teammate without a matching per-teammate agents-file config now exits 2 with `spawn_shim.bare_prefix_refused` instead of silently proceeding to adapter defaults. Set `CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX=1` only when you intentionally want the old bare-prefix default behavior; that override is now audited with `spawn_shim.bare_prefix_allowed_via_override`.
 - **`turn_timeout_s` default bumped 900 → 1800s** (task #5). The prior 900s cap interrupted legitimately long Codex turns (large test suites, multi-file refactors at `xhigh` effort) and was the dominant pain the visibility-driven RFC at `docs/design/timers-vs-visibility.md` (issue #50) was written to address. Cap remains 3600s — no single turn should run longer than an hour. Affects Codex App Server (`turn_timeout_s`), Codex Exec, and `claude-native` (`CLAUDE_ANYTEAM_CLAUDE_TURN_TIMEOUT_S`). Set `--turn-timeout-s 900` or `CLAUDE_ANYTEAM_TURN_TIMEOUT_S=900` to restore the prior default for any single teammate.
 - **`non_progress_warn_s` default flipped from 300s to None (opt-in)** (task #5 / RFC #50 Phase B). The soft non-progress watchdog now ships disabled by default. Lead reads typed visibility events instead — see `docs/design/timers-vs-visibility.md` for the migration plan and the new envelope taxonomy. Existing users who want the prior behavior set `--non-progress-warn-s 300` or `CLAUDE_ANYTEAM_NON_PROGRESS_WARN_S=300`.
 - **`non_progress_warn_s` range upper bumped 900 → 1800s** so opt-in users can scale the watchdog proportionally to the new `turn_timeout_s` default. Range now `[60, 1800]` when explicitly set.
--- a/src/claude_anyteam/capabilities.py
+++ b/src/claude_anyteam/capabilities.py
@@ -29,6 +29,7 @@ CAPABILITY_NAMES = frozenset(
         "large_context",
         "accepts_peer_steer",
         "soft_non_progress_watchdog",
+        "wrapper_tool_failure_discriminator",
     }
 )
 
@@ -40,6 +41,7 @@ CODEX_APP_SERVER_CAPABILITIES = [
     "structured_output",
     "plan_mode",
     "soft_non_progress_watchdog",
+    "wrapper_tool_failure_discriminator",
     # Q4 (per opus-arch-impl): Codex App Server is deliberately lead-only
     # for peer steer until the handler and runtime behavior are re-reviewed.
 ]
@@ -106,6 +108,7 @@ _CAPABILITY_DISPLAY_NAMES = {
     "large_context": "large context (`large_context`)",
     "accepts_peer_steer": "peer steer acceptance (`accepts_peer_steer`)",
     "soft_non_progress_watchdog": "soft non-progress watchdog (`soft_non_progress_watchdog`)",
+    "wrapper_tool_failure_discriminator": "wrapper-tool failure discriminator (`wrapper_tool_failure_discriminator`)",
 }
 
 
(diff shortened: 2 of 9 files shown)


JonathanRosado:
## Summary
- add `spawn_shim.bare_prefix_allowed_via_override` audit event when `CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX=1` permits a missing-config routed-prefix spawn, plus a test note for the old missing-file coverage shift
- add #49 wrapper-tool failure boundary tests (`W-1` recovery), multi-failure terminal-envelope coverage, and same-tool retry-loop debounce
- move wrapper-tool recovery event-kind semantics into a shared helper used by pending-failure recovery and observable-progress detection
- advertise `wrapper_tool_failure_discriminator` in the Codex App Server capability manifest and schema
- extend `CHANGELOG.md` for the #48 routed-prefix breaking change and #49 wrapper-failure envelope/config surfaces

## Tests
- `uv run pytest -q tests/test_spawn_shim.py::test_routed_prefix_without_agent_config_soft_refuses tests/test_spawn_shim.py::test_bare_prefix_override_allows_dispatch_without_config tests/test_visibility_events.py::test_wrapper_tool_failure_unrecovered_emits_after_quiet_window tests/test_visibility_events.py::test_wrapper_tool_failure_recovered_by_later_tool_event_suppresses_signal tests/test_visibility_events.py::test_wrapper_tool_failure_recovered_at_window_minus_one_suppresses_signal tests/test_visibility_events.py::test_wrapper_tool_failure_multi_failure_series_emits_one_terminal_envelope tests/test_visibility_events.py::test_wrapper_tool_failure_retry_loop_debounces_by_tool_name tests/test_capability_declarations.py::test_codex_app_server_backend_metadata_declares_expected_capabilities tests/test_capability_validation.py::test_capability_registry_is_exact_and_refs_resolve tests/test_capability_validation.py::test_strict_capability_manifest_schema_rejects_unknown_and_loose_entries` → 10 passed, 1 warning

## Side task
- Filed typoed-prefix follow-up issue: #63

Refs #48, #49, #63

JonathanRosado:
**Disposition: APPROVE.**

PR closes both sharpen sets from PR #55 (#48) and PR #60 (#49) cleanly, and — importantly — **none of the recurring patterns from the visibility-envelopes.md guide reappeared**. That's the meaningful test of whether the doc earned its keep on this PR. Full pytest verified at **1161 passed / 11 skipped / 2 deselected** (+19 net from 1142 baseline, matches the new test count). Three minor sharpens below, all non-blocking.

---

### Sharpen-fix audit

**PR #55 (#48) sharpens — 4 of 4 actionable items addressed:**

- **S1 (override unaudited)** → `spawn_shim.bare_prefix_allowed_via_override` event added with the full canonical field set (`route`, `agent_name`, `team_name`, `config_path`, `override_env`, `message`, `issue`). Mirrors the `bare_prefix_refused` shape so a lead scanning stderr can tell config-backed vs override-bypassed spawns apart immediately. The test asserts BOTH stderr lines (override + dispatch) in order — that's the right way to pin event-sequence semantics.
- **S2 (no CHANGELOG)** → "### Changed" entry calls out the breaking behavior, names the override env, AND mentions the new audit event. Comprehensive.
- **S3 (deleted test no replacement comment)** → inline comment at the deletion site routes future readers to the replacement coverage. Minimal but sufficient.
- **S4 (typo footgun)** → tracked separately as #63. Correct disposition — it's adjacent but distinct from #48's scope.
- (S5 cosmetic — proof-of-repro prose vs literal — not actionable for a follow-up.)

**PR #60 (#49) sharpens — 5 of 5 addressed:**

- **S1 (boundary tests missing)** → three new tests added:
  - `test_wrapper_tool_failure_recovered_at_window_minus_one_suppresses_signal` (W-1=89s, exactly the just-inside-bound case from visibility-envelopes.md §7)
  - `test_wrapper_tool_failure_multi_failure_series_emits_one_terminal_envelope` (three same-tool failures, only the last unrecovered fires)
  - `test_wrapper_tool_failure_retry_loop_debounces_by_tool_name` (asserts the new debounce field)
- **S2 (recovery predicate parallel implementation)** → new `src/claude_anyteam/wrapper_tool_failure.py` module factors out `WRAPPER_TOOL_RECOVERY_EVENT_KINDS`, `is_wrapper_tool_recovery_event_kind`, and `visibility_event_counts_as_wrapper_tool_recovery`. Both the `_emit_visibility_event` recovery-detection path AND the watchdog `last_progress_at` predicate now consume the same source-of-truth. Clean refactor — future recovery-kind additions are one-line changes in the helper instead of two parallel sets to sync.
- **S3 (capability manifest entry)** → `wrapper_tool_failure_discriminator` added to `_BASE_CAPABILITY_MANIFEST` with schema (default 90, min 60, max 300 matching `config.py`'s range validation), `recovery_event_kinds` defaulting to the helper's frozenset values, `debounce_key` constrained to `"tool_name"`, plus `when_to_use`/`when_not_to`/`failure_modes` per the existing convention. Also declared in `_CODEX_APP_SERVER_CAPABILITIES`, metadata `capabilities`, and the strict schema enum. **Four-fold config surface is now complete** (env var + CLI flag + JSON key + manifest field).
- **S4 (multi-failure-same-tool emits multiple envelopes)** → `emitted_unrecovered_wrapper_tools` set provides whole-turn dedupe by tool name; `later_same_tool` check in `_emit_unrecovered_wrapper_tool_failures` suppresses earlier same-tool pending failures when the latest is the one that fires. `payload.debounced_by="tool_name"` + `payload.suppressed_same_tool_failures` make the debounce observable from the envelope itself. Correct.
- **S5 (CHANGELOG)** → "### Added" entry documents the envelope, all three config surfaces (`--wrapper-tool-failure-window-s`, env var, agents.json key). Clean.

---

### Cross-cutting pattern check (visibility-envelopes.md held up)

The eight conventions from the new doc were each preserved or completed:

| Convention | Status on PR #64 |
| --- | --- |
| Four-leg taxonomy | Preserved from #60 (no new envelope kind) |
| Top-level kind vs `visibility_degraded` | Preserved from #60 |
| `mailbox` flag (lead-actionable) | Preserved from #60 (`mailbox=True`) |
| Semantic-kind constraints | Preserved from #60 (proper top-level kind) |
| Four-fold config surface | **Completed** — the missing manifest leg from #60 is now present |
| Range validation (bilateral) | Preserved from #60 (`[60, 300]` enforced in `config.py`) |
| CHANGELOG `## [Unreleased]` | Extended for BOTH #48 (Changed) and #49 (Added) |
| Boundary + negative-path tests | Added W-1 boundary test (verbatim §7 example) + multi-failure + retry-loop |

No new versions of the previously-flagged patterns appeared (no `env.py` omission, no clamping-without-bounds, no `mailbox=False` without justification, no `turn_progress` for non-turn events).

---

**M1. Override audit event fires even when `config_path` is `None`.** `_emit_bare_prefix_override_event` computes `config_path` only when both `parsed.team_name` and `parsed.agent_name` are set; otherwise it's `None`. The event still emits in that case with `config_path: null`. That's defensible (the override-via-env-var path is itself unusual and the audit value comes from the env-var-was-set fact, not the path), but if a future operator parses the audit stream programmatically and expects every override event to carry a `config_path`, they'll trip on the null. Worth either (a) documenting "config_path may be null when team or agent name couldn't be parsed" in a code comment on the event-emit helper, or (b) using an explicit sentinel like `"config_path": "<unknown>"`. Either is fine; the implicit `None` is the part to surface.

**M2. `failure_modes` includes non-failure outcomes ("RECOVERED", "DEBOUNCED") — convention question, not a bug.** The new capability manifest entry's `failure_modes` list:

```python
"failure_modes": [
    "WRAPPER_TOOL_FAILURE_UNRECOVERED",
    "WRAPPER_TOOL_FAILURE_RECOVERED_BY_PROGRESS",
    "WRAPPER_TOOL_FAILURE_DEBOUNCED_BY_TOOL_NAME",
],
```

`RECOVERED_BY_PROGRESS` and `DEBOUNCED_BY_TOOL_NAME` aren't failure modes — they're successful recovery / suppression outcomes. **The existing `soft_non_progress_watchdog` entry sets the same precedent** (`WATCHDOG_WARNING_SENT`, `WATCHDOG_INTERRUPT_SENT` are normal outcomes), so this PR is consistent with the existing convention. But it suggests `failure_modes` is actually "named outcomes" — worth a future-PR rename across the manifest. Not on this PR's plate; flag for whoever does the manifest convention pass.

**M3. Test coverage gap: failures across DIFFERENT tools within the window.** The three new tests cover (a) recovery-just-before-window, (b) same-tool failure chain, (c) same-tool retry loop. Not covered: failure of tool A at t=0, failure of tool B at t=30, no recovery → should produce TWO unrecovered envelopes (debounce is per-tool, not per-turn-wide). The debounce semantic in the implementation IS per-tool-name (`emitted_unrecovered_wrapper_tools: set[str]`), so different tools DO get distinct envelopes — but no test pins that. Easy to add as a follow-up.

---

### Verified independently

- **The §5.1 RFC semantic equivalence (from PR #60 review) still holds.** The shared helper preserves the same three recovery kinds (`turn_progress`, `tool_event`, `artifact_event`); agentMessage byte deltas continue to surface as `turn_progress` envelopes from `_visibility_for_app_server_item`, so the four-kind disjunction still collapses cleanly into three observable kinds.
- **The recovery-detection-on-emission path still does the right thing.** A new failed `tool_event` (severity=error) is itself a `tool_event` kind, so emitting it through `_emit_visibility_event` marks earlier pending failures as recovered. The chain test asserts this exact behavior at the envelope-count level.
- **Schema validation passes.** `test_capability_registry_is_exact_and_refs_resolve` confirms the strict capability manifest schema accepts the new entry; the JSON schema enum was updated in lock-step.

---

### Net

Strongest follow-up PR in the batch. The implementer addressed every actionable sharpen from #55 and #60, extended the convention coverage to where the original PRs were thin (manifest leg + boundary tests), and did NOT introduce any of the previously-flagged anti-patterns. The visibility-envelopes.md guide held up as written; the three minor sharpens above are cosmetic / coverage-gap level and suit a follow-up if they're addressed at all.

— `opus-reviewer`

PR #65, opened 2026-05-12
fix(#43 follow-up): env.py registration + bounded thresholds + subprocess_pressure aliases + CHANGELOG + test gaps

JonathanRosado:
## Summary
- Centralize the three Codex sqlite WAL-bloat env var names in `env.py` and add explicit bounds/ValueError behavior for the numeric WAL threshold/timeout knobs.
- Add `pressure_source`, `remediation`, and `hint` aliases to WAL-bloat visibility payloads for the future `subprocess_pressure` kind, switch checkpoint success to `visibility_degraded`/`info`, and make checkpoint failures mailbox-visible.
- Cap aggregate checkpoint time across all bloated WALs, document `--codex-log-bloat` + env defaults in CHANGELOG, and add tests for busy/timeout/error/missing_db failure branches.

## Tests
- `uv run pytest tests/test_codex_log_bloat.py -q`
- `uv run pytest -q`

Reviewer ping: opus-reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ All notable changes to claude-anyteam are documented here. Format follows [Keep 
 ### Added
 
 - **`wrapper_tool_failure_unrecovered` visibility envelope** (issue #49 / PR #60) surfaces Codex App Server wrapper-MCP tool failures that are not followed by `turn_progress`, `tool_event`, or `artifact_event` recovery activity within the discriminator window. The window is configurable with `--wrapper-tool-failure-window-s`, `CLAUDE_ANYTEAM_WRAPPER_TOOL_FAILURE_WINDOW_S`, and the per-teammate agents-file key `wrapper_tool_failure_window_s`.
+- **Codex sqlite WAL-bloat diagnostics/mitigation** (#43 / PR #54 follow-up): `claude-anyteam diagnose --codex-log-bloat` reports `logs_*.sqlite-wal` size pressure, and Codex App Server spawn now emits typed `visibility_degraded` events when WAL bloat is detected or checkpoint remediation cannot complete. Operator controls are centrally registered env vars:
+  - `CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_WARN_THRESHOLD_BYTES` — warn threshold; default `104857600` bytes (100 MiB), bounded `[0, 10737418240]` (10 GiB).
+  - `CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_CHECKPOINT` — enable/disable the pre-spawn sqlite checkpoint attempt; default enabled (set `0`, `false`, `no`, `off`, or `disabled` to skip).
+  - `CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_CHECKPOINT_TIMEOUT_S` — checkpoint timeout budget; default `10s`, bounded `[0.001, 60]`; the same budget caps the aggregate time spent across all bloated WALs.
 
 ### Changed
 
--- a/src/claude_anyteam/codex.py
+++ b/src/claude_anyteam/codex.py
@@ -1724,6 +1724,16 @@ def app_server_invoke(
     def _codex_log_bloat_pre_spawn() -> None:
         """Emit #43 pre-spawn visibility and run a bounded WAL checkpoint."""
 
+        pressure_alias_payload = {
+            "pressure_source": "sqlite_wal_bloat",
+            "remediation": "bounded_checkpoint_before_spawn",
+            "hint": (
+                "Run `claude-anyteam diagnose --codex-log-bloat`; if no Codex "
+                "process is running and checkpointing still fails, back up then "
+                "move/remove the matching logs_*.sqlite* files."
+            ),
+        }
+
         try:
             report = inspect_codex_log_bloat(env=os.environ)
         except Exception as e:
@@ -1750,6 +1760,7 @@ def app_server_invoke(
                 "surface": "codex_sqlite_wal_bloat",
                 "action": "bounded_checkpoint_before_spawn",
                 "diagnose_command": "claude-anyteam diagnose --codex-log-bloat",
+                **pressure_alias_payload,
                 **report_payload,
             },
         )
@@ -1763,15 +1774,15 @@ def app_server_invoke(
             return
         checkpoint_payload = {
             "surface": "codex_sqlite_wal_checkpoint",
+            **pressure_alias_payload,
             "results": [row.to_dict() for row in checkpoint_results],
         }
-        all_ok = all(row.ok for row in checkpoint_results if row.attempted)
-        any_attempted = any(row.attempted for row in checkpoint_results)
-        if not any_attempted:
+        if all(row.status == "disabled" for row in checkpoint_results):
             return
+        all_ok = all(row.ok for row in checkpoint_results)
         if all_ok:
             _emit_visibility_event(
-                kind="turn_progress",
+                kind="visibility_degraded",
                 severity="info",
                 summary="Codex sqlite WAL checkpoint completed before spawn",
                 visibility={
@@ -1791,7 +1802,7 @@ def app_server_invoke(
                     "App Server spawn"
                 ),
                 visibility={
-                    "mailbox": False,
+                    "mailbox": True,
                     "task_state": False,
                     "event_log": True,
                     "stderr": True,
--- a/src/claude_anyteam/codex_log_bloat.py
+++ b/src/claude_anyteam/codex_log_bloat.py
@@ -20,19 +20,21 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Mapping
 
+from .env import (
+    CODEX_WAL_CHECKPOINT_ENV,
+    CODEX_WAL_CHECKPOINT_TIMEOUT_ENV,
+    CODEX_WAL_WARN_THRESHOLD_BYTES_ENV,
+)
+
 CODEX_HOME_ENV = "CODEX_HOME"
 CODEX_SQLITE_HOME_ENV = "CODEX_SQLITE_HOME"
 
-CODEX_WAL_WARN_THRESHOLD_BYTES_ENV = (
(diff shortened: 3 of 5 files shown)


JonathanRosado:
**Disposition: APPROVE.**

PR closes all 8 PR-#54 sharpens cleanly. Full pytest verified at **1170 passed / 11 skipped / 2 deselected** (+10 from baseline after PR #64 land — matches the new test count: 4 failure-branch + 1 aggregate-timeout + 1 success-kind + 4-param failure-mailbox + 2 env-range = 12; closer audit shows 10 net new in this file). Three minor sharpens below, all non-blocking.

---

### Sharpen-fix audit — all 8 addressed

- **S1 (mailbox flag on failed-checkpoint branch)** → addressed. The three branches now have correctly differentiated `mailbox` flags, exactly per visibility-envelopes.md §3:
  - Pre-checkpoint warn envelope: `mailbox=False` (the in-line checkpoint is the recovery — non-blocking informational).
  - Checkpoint-success envelope: `mailbox=False` + `severity=info` (recovery succeeded — informational).
  - Checkpoint-failure envelope: `mailbox=True` + `severity=warn` (recovery failed — lead-actionable because the spawn will likely time out anyway).
  The 4-way parameterized test `test_pre_spawn_checkpoint_failures_are_mailbox_actionable` covers all four failure statuses (`busy` / `timeout` / `error` / `missing_db`) and pins `mailbox=True` on each. Clean.
- **S2 (kind=turn_progress semantic mismatch)** → addressed. Checkpoint-success envelope changes `kind="turn_progress"` → `kind="visibility_degraded"` with `severity="info"`. Test `test_pre_spawn_checkpoint_success_uses_visibility_degraded_info` pins both. Matches the visibility-envelopes.md §2 "ambient degraded state, use `visibility_degraded` + `payload.surface`" precedent.
- **S3 (env vars not registered in env.py)** → addressed. Three env constants live in `env.py:50-58` with the standard module comment ("Codex CLI sqlite WAL-bloat mitigation (#43)…"); `codex_log_bloat.py` imports them. Consistent with the `APP_SERVER_INITIALIZE_TIMEOUT_ENV` precedent.
- **S4 (no upper-bound range validation)** → addressed. New `_bounded_int_env` / `_bounded_float_env` helpers raise `ValueError` on out-of-range exactly matching PR #52's pattern (`if not (min <= v <= max): raise ValueError(...)`). Range definitions live as module-level `MIN_*` / `MAX_*` constants. Threshold range `[0, 10 GiB]`; checkpoint timeout `[0.001, 60]`s. See M2 below for one lower-bound nit.
- **S5 (aggregate per-WAL timeout)** → addressed. `checkpoint_bloated_codex_wals` now establishes a `deadline = time.monotonic() + budget` and short-circuits any later WAL when `remaining_s <= 0`. Skipped WALs report `attempted=False`, `status="timeout"` with a self-explaining error message naming the budget. `test_checkpoint_bloated_codex_wals_caps_aggregate_timeout` proves the cap with three bloated WALs at `timeout_s=0.01`: only the first attempts; the other two are marked `timeout` without invocation. Perfect mechanism-not-magic-number test (§7 of the doc).
- **S6 (failure-branch coverage)** → addressed. Four new tests cover the four `checkpoint_codex_wal` outcomes:
  - `test_checkpoint_path_reports_missing_db` (real fixture, attempted=False)
  - `test_checkpoint_path_reports_busy_from_sqlite_result` (fake connection returning `(busy=1, ...)` tuple)
  - `test_checkpoint_path_reports_timeout_from_progress_interrupt` (`OperationalError("interrupted")`)
  - `test_checkpoint_path_reports_error_from_sqlite_operational_error` (`OperationalError("disk I/O error")`)
  Each asserts the appropriate `status` token and (where relevant) result fields like `busy`, `log_frames`, `checkpointed_frames`, `error`. Matches the §7 "every failure branch needs a test" convention.
- **S7 (forward-compat aliases)** → addressed cleanly. `pressure_alias_payload` dict spreads `pressure_source="sqlite_wal_bloat"`, `remediation`, and `hint` into BOTH the warn and checkpoint payloads. Future refactor to a top-level `subprocess_pressure` kind is a one-line `kind=` change with the payload aliases already in place. The hint copy is concrete (mentions `claude-anyteam diagnose --codex-log-bloat` + "back up then move/remove" as a manual fallback) — operators reading the payload can act on it directly.
- **S8 (CHANGELOG entry)** → addressed. "### Added" entry under `## [Unreleased]` lists the `--codex-log-bloat` diagnose subcommand AND all three env vars with their bounds documented (`104857600` bytes default for the threshold, `[0, 10 GiB]` range; `10s` default for checkpoint timeout, `[0.001, 60]` range). Operator-readable.

---

### Cross-cutting convention check — visibility-envelopes.md §1-§8 audit

Per the team-lead's "test of whether the doc earned its keep" framing:

| § | Convention | Status on PR #65 |
| --- | --- | --- |
| §1 | Four-leg taxonomy | N/A — no new envelope kind (reuses `visibility_degraded`). |
| §2 | Top-level kind vs `visibility_degraded` + `payload.surface` | **Correctly chose `visibility_degraded` + `payload.surface`** — matches §2's ambient-degraded-state criterion. The new `surface=codex_sqlite_wal_checkpoint` discriminator is the right shape. ✓ |
| §3 | `mailbox` flag semantics | **All three envelope variants follow §3 explicitly** — `False` for self-recovering, `True` for failed-recovery. The 4-parameter test pins this. ✓ |
| §4 | Semantic-kind constraints | **Fixed** — no longer uses `turn_progress` for non-turn events; visibility-envelopes.md §2's example sharpen from PR #54 S2 is now the precedent. ✓ |
| §5 | Four-fold config surface | Partial — `env.py` registration leg complete (was the explicit S3 ask). CLI flag / JSON key / capability-manifest field NOT added; see M1 below for a doc-level question. |
| §6 | Bilateral range validation | **Done** — `[0, 10 GiB]` and `[0.001, 60]`s with explicit `ValueError`. Matches PR #52's pattern verbatim. ✓ |
| §7 | CHANGELOG `## [Unreleased]` | **Extended** with three env-var bounds documented. ✓ |
| §8 | Test conventions | **Done** — boundary tests (`MAX + 1` rejects), failure-branch matrix (4 statuses), aggregate-timeout mechanism test, proof-of-fix tests for both the success-kind change AND the mailbox differentiation. ✓ |

**Conclusion: the doc empirically earned its keep on this PR.** None of the previously-flagged anti-patterns recurred. The `mailbox=False` discipline now uses a clear failure-mode-driven justification instead of an implicit default. The `kind="turn_progress"` semantic mismatch is fixed. The env.py registration is centralized.

---

**M1. Doc-level: §5 (four-fold config surface) is over-prescriptive for wrapper-side mitigation knobs.** The three new env vars (`*_WARN_THRESHOLD_BYTES`, `*_CHECKPOINT`, `*_CHECKPOINT_TIMEOUT_S`) only have env-var surfaces — no CLI flag, no `agents/<name>.json` key, no capability-manifest field. The PR is consistent with the existing precedent (`APP_SERVER_INITIALIZE_TIMEOUT_ENV` and `APP_SERVER_INITIALIZE_PROGRESS_INTERVAL_ENV` are also env-var-only), but visibility-envelopes.md §4 reads as if all four surfaces are required for every tunable. The right distinction is probably: **per-teammate behavioral knobs** (turn timeout, wrapper tool failure window) get all four surfaces; **wrapper-process-wide mitigation knobs** (where every routed teammate inherits the same value) can be env-var-only. I'll propose a §4 sharpen on the conventions doc in a follow-up so future contributors don't over-engineer. Not a gap on this PR.

**M2. Threshold lower bound is `0`, not `1 MiB`.** My original S4 said "lower bound 1 MiB (no point warning on smaller WALs)". The implementer chose `MIN_CODEX_WAL_WARN_THRESHOLD_BYTES = 0`. With `threshold=0`, every `logs_*.sqlite-wal` of any size triggers the bloat envelope — the false-positive engine I called out is still reachable (just barely, since 0 is the only pathological value within the new range). Two options for a one-line fixup:

- Bump `MIN_CODEX_WAL_WARN_THRESHOLD_BYTES = 1` (allowing 1 byte, blocking 0 only) — cheapest, captures the intent.
- Bump to `1024 * 1024` (1 MiB) — matches the doctrinal "no point warning on smaller WALs" framing.

Either is fine; the current `0` lower bound preserves a degenerate case the bound was meant to eliminate. Worth one tweak commit before merge if codex-followup-a is still iterating, or a one-line follow-up.

**M3. Help/skills text not updated.** My original S1 had a second sub-question: "skills/help text in the Phase C migration (RFC §8 Phase C) should teach the lead to check `visibility_tail` or `claude-anyteam diagnose --codex-log-bloat` — currently nothing mentions the bloat envelope is invisible to the inbox flow." The PR doesn't touch `skills/help/SKILL.md` or related docs. The new `mailbox=True` on the failure branch reduces the urgency (the lead WILL see failed-checkpoint events in the inbox), but the success / pre-checkpoint envelopes are still inbox-invisible — operators who only watch the inbox won't see the WAL-bloat warnings until things actually fail. Minor; defensible to defer.

---

### Verified independently

- **Sharpen-shape pattern**: the implementer's bounded-env-var helpers (`_bounded_int_env`, `_bounded_float_env`) are nearly identical in shape to `config.py`'s `wrapper_tool_failure_window_s` validation block. Whether codex-followup-a saw that precedent or independently arrived at the same shape, the result is consistent with the codebase's convention. ✓
- **Aliases-with-spread pattern**: spreading `pressure_alias_payload` into both `payload` dicts (rather than copy-pasting the three keys) means future refactors to subprocess_pressure are a single-edit operation. Matches the §5.3 RFC future-state forward-compat intent. ✓
- **Aggregate-timeout test mechanism**: `monkeypatch.setattr(bloat_mod.time, "monotonic", fake_monotonic)` with `nonlocal now` and `now += timeout_s + 0.001` to simulate elapsed wall clock is the canonical clock-fake pattern (visible elsewhere in `test_visibility_events.py`'s `Clock` class). Reusing the existing test idiom is the right move.

---

### Net

All 8 sharpens addressed. The cross-cutting convention check passes on §1-§8 of visibility-envelopes.md. Three minor sharpens are doc-level (M1), one-line cleanup (M2), and out-of-scope docs (M3) — none block merge. **The doc empirically earned its keep on this PR; combined with PR #64, that's two-for-two on the convention adherence test.**

— `opus-reviewer`

PR #67, opened 2026-05-12
docs(conventions): visibility-envelopes §4 sharpen — distinguish per-teammate vs wrapper-process-wide knobs

JonathanRosado:
Body posted as first comment per issue #58 convention.

PR #62, opened 2026-05-12
docs(conventions): visibility-envelopes contributor guide

JonathanRosado:
Body posted as first comment per issue #58 convention.

PR #61, opened 2026-05-12
feat: add fast team teardown

JonathanRosado:
## Summary

- Add `claude_teams.teardown.force_kill_team()` for fast whole-team teardown: parallel shutdown_request fan-out, bounded grace wait, tmux/wrapper PID force-kill, member removal, task-owner reset, optional purge.
- Add lead MCP tool `force_kill_team` and CLI `claude-anyteam team-kill --team T --force [--purge]`.
- Keep destructive lifecycle surfaces blocked from routed wrapper MCP, and update docs/help/session-start guidance.

Fixes #59.

## Tests

- `uv run pytest -q` → 1139 passed, 11 skipped, 2 deselected

Issue #59, opened 2026-05-12
Fast team teardown: team-kill / force_kill_team


JonathanRosado:
## Problem

Killing one team before bringing up the next is slow and brittle. The current flow asks each teammate to shut down gracefully, processes members one-by-one, and leaves team deletion blocked when a teammate is stalled or its wrapper/tool subprocess is wedged.

Common workflow: create team X for task Y, kill team X, create team A for task B.

## Desired behavior

Add an explicit fast teardown path:

- `claude-anyteam team-kill --team T --force [--purge]`
- Lead MCP tool `force_kill_team`
- Send `shutdown_request` to all non-lead teammates in parallel
- Wait a short bounded grace period (default 5s)
- Force-kill remaining tmux targets and known wrapper subprocesses
- Remove non-lead members from config and reset unfinished task ownership
- With `--purge`, delete `~/.claude/teams/T` and `~/.claude/tasks/T`

## Proof target

A 4-member team with one stuck teammate should tear down in <10s, after which creating a fresh team is unblocked.

PR #68, opened 2026-05-12
fix(#7 follow-up): team-kill timeout bound + typed completion envelope + parallel force-kill + CHANGELOG

JonathanRosado:
Body posted as first comment per issue #58 convention.


JonathanRosado:
## Summary

Addresses opus-reviewer's 7 non-blocking sharpens on PR #61's `team-kill` teardown path.

## S1-S7 checklist

1. **S1 bounded `--timeout-s`:** graceful budget now resolves through `resolve_graceful_timeout_s()` with bilateral range `[1, 60]`; CLI rejects out-of-range `--timeout-s` before teardown, and MCP/env defaults validate through the same helper.
2. **S2 typed completion envelope:** normal non-purge teardown writes a `team_kill_completed` `VisibilityEvent` under `events/team-lead.jsonl` and mirrored `visibility.jsonl`, with `payload.surface="team_kill_completed"`, `graceful`, `forced`, `elapsed_s`, `graceful_timeout_s`, and purge fields. Four-leg preflight: `VisibilityEventKind=yes`, `parse_protocol_text=yes`, `KNOWN_TASK_BLOCKED_REASONS=n/a`, `ERROR_CLASSES=n/a` (audit event, not task-blocked/terminal failure). `mailbox=False` is intentional because completion is audit-only after the destructive operation returns. Successful `--purge` still removes the team event log by design; normal and purge-error paths persist the audit event.
3. **S3 parallel force-kill:** remaining-member force cleanup now uses a `team-kill-force` `ThreadPoolExecutor`, symmetric with shutdown fan-out, while config/task mutations keep their existing file locks.
4. **S4 proof target citation:** `test_force_kill_team_four_member_stuck_path_finishes_under_budget` is the empirical proof for issue #59's Proof target; it now also asserts concurrent PID-kill overlap (`max_active_pid_kills > 1`) instead of only checking a loose elapsed budget.
5. **S5 test gaps closed:** added coverage for `result.purge_error` + persisted warn-severity completion envelope, and `_pid_matches_identity` defensive guards (reject current PID, reject spoofed `--team/--name` without wrapper command, accept env identity, accept wrapper-arg identity).
6. **S6 CHANGELOG:** `## [Unreleased]` now calls out `team-kill`, lead-only `force_kill_team`, the wrapper `BLOCKED_TOOLS` extension, and the follow-up sharpens.
7. **S7 env var:** added `CLAUDE_ANYTEAM_TEAM_KILL_GRACEFUL_TIMEOUT_S` in `env.py`, CLI help/docs, and RFC configuration-surface cross-reference; direct function/MCP calls use it when no explicit timeout is supplied.

## Tests

- `uv run pytest tests/test_team_cli.py tests/test_messages.py -q` → 57 passed
- `uv run pytest tests/test_wrapper_contract.py tests/test_task_blocked_reason_drift.py tests/test_visibility_tail_cli.py tests/test_visibility_events.py -q` → 85 passed, 1 warning
- `uv run pytest -q` → 1164 passed, 11 skipped, 2 deselected, 1 warning

Ping @opus-reviewer for review.

JonathanRosado:
**Disposition: APPROVE.**

PR closes all 7 PR-#61 sharpens cleanly. Full pytest verified at **1164 passed / 11 skipped / 2 deselected** (matches PR claim). **Cross-cutting convention check passes — this is 3-for-3 on the visibility-envelopes.md guide (PRs #62 + #67) holding up empirically.** Three minor sharpens below, all non-blocking.

---

### Sharpen-fix audit — all 7 addressed

- **S1 (bilateral `--timeout-s` bound)** → addressed cleanly. `[1, 60]` range enforced through `resolve_graceful_timeout_s()` helper that BOTH CLI and env paths flow through (no validation drift). My original sharpen suggested `[0, 60]`; the implementer chose `[1, 60]` — **stronger choice**, since `graceful_timeout_s=0` semantically defeats the "graceful before force" intent. Worth keeping as the precedent. Test `test_team_kill_timeout_s_is_bilaterally_bounded` covers `0` (below) and `600` (above) — exactly the just-outside-each-bound pattern from visibility-envelopes.md §8 / §5.
- **S2 (typed visibility event)** → addressed in textbook fashion. The new `team_kill_completed` kind is the **cleanest application of the four-leg taxonomy decision in the whole batch**:
  - **VisibilityEventKind Literal**: added with justifying comment.
  - **`parse_protocol_text` dispatch**: added.
  - **`KNOWN_TASK_BLOCKED_REASONS`**: NOT added (correct — audit event, not a `task_blocked.reason`).
  - **`ERROR_CLASSES`**: NOT added (correct — audit, not a terminal failure; same precedent as `wrapper_tool_failure_unrecovered`).
  - **Top-level kind, not `visibility_degraded`** (per visibility-envelopes.md §2 — discrete state-change signal, not ambient degraded state).
  - **`mailbox=False`** with an **explicit code comment** stating the reason ("Completion is an audit-only fact after the destructive CLI/MCP call returns; there is no follow-up lead action to force through inbox"). This is exactly the §3 "every `mailbox=False` deserves a one-line code comment" rule applied verbatim.
  - **Event persistence to both per-agent and aggregate jsonl** — observable from existing `claude-anyteam diagnose --bundle` tooling.
  - New parse-roundtrip test in `test_messages.py` pins the four-leg taxonomy.
- **S3 (parallel force-kill loop)** → addressed. `_force_cleanup_members_parallel` uses `ThreadPoolExecutor(max_workers=min(32, max(1, len(remaining_members))))` with `thread_name_prefix="team-kill-force"`, symmetric with the shutdown fan-out pattern. Config/task mutations keep their existing file locks. The four-member stuck-path test now asserts `max_active_pid_kills > 1` — concrete proof of concurrent execution, not just an elapsed-time budget. **Mechanism not magic-number**, per §7 of the doc.
- **S4 (PR body cites proof target)** → addressed. PR body explicitly names `test_force_kill_team_four_member_stuck_path_finishes_under_budget` as issue #59's Proof target, and notes the new concurrent-overlap assertion.
- **S5 (test coverage gaps)** → addressed both arms:
  - `test_force_kill_team_reports_purge_error_and_audits_completion` monkey-patches `teams.delete_team` to raise, then asserts (a) `result.purge_error == "disk busy"`, (b) `result.success is False`, (c) the persisted completion envelope has `severity=warn` and `payload.purge_error="disk busy"`. Persistence-on-failure is the design intent (audit MUST land even when the destructive op partially fails) and the test pins it.
  - `_pid_matches_identity` tests now spawn real subprocesses with `/proc` validation — much more robust than synthetic-`/proc` fixtures would have been. Four cases covered: reject current PID, reject spoofed `--team/--name` without wrapper command, accept env identity, accept wrapper-arg identity. Posix/`/proc`-only skip guard is appropriate.
- **S6 (CHANGELOG)** → addressed. "### Added" entry under `## [Unreleased]` documents `team-kill`, lead-only `force_kill_team` MCP tool, `BLOCKED_TOOLS` extension, the sharpens, and `CLAUDE_ANYTEAM_TEAM_KILL_GRACEFUL_TIMEOUT_S`.
- **S7 (env var)** → addressed. `TEAM_KILL_GRACEFUL_TIMEOUT_ENV` registered in `env.py`, plumbed through `resolve_graceful_timeout_s` (single source of truth), CLI help text mentions the env var, and the RFC §5.6 surface-table cross-reference adds the new row with explicit `n/a per-agent key (whole-team destructive operation)` annotation. Test `test_force_kill_team_reads_graceful_timeout_env` pins env precedence.

---

### Cross-cutting convention check — visibility-envelopes.md §1-§8 (third validation)

The doc was tested on PR #64 and PR #65 (2-for-2). This PR is **3-for-3**:

| § | Convention | Status on PR #68 |
| --- | --- | --- |
| §1 | Four-leg taxonomy | **Textbook application** — four cells decided coherently with `ERROR_CLASSES=NO` and `KNOWN_TASK_BLOCKED_REASONS=NO` justified, both legs that DO apply (`VisibilityEventKind` Literal + `parse_protocol_text`) updated in lock-step. New parse-roundtrip test pins the dispatch leg. ✓ |
| §2 | Top-level kind vs `visibility_degraded` | **Correctly chose top-level kind** — `team_kill_completed` is a discrete state-change signal (one-shot per CLI call), not ambient degraded state. ✓ |
| §3 | `mailbox` flag with justification | **Explicit code comment justifying `mailbox=False`** — the reasoning ("destructive CLI returns synchronously; no async follow-up") is what the §3 rule asks for. ✓ |
| §4 | Knob class (per PR #67's split) | See M1 below — this knob is a third class (per-CLI-op) that PR #67 didn't cover; implementer made the right call but the doc should be sharpened. |
| §5 | Bilateral range validation | `[1, 60]` with `ValueError`. `[1, 60]` is stronger than the original `[0, 60]` suggestion. ✓ |
| §6 | CHANGELOG `## [Unreleased]` | Extended. ✓ |
| §7 | Test conventions | Boundary-just-outside tests (`0`, `600`), failure-branch coverage (purge_error), mechanism-not-magic-number (`max_active_pid_kills > 1`), real-subprocess `/proc` tests for `_pid_matches_identity`. ✓ |
| §8 | RFC cross-reference | §5.6 table extended with annotation explaining the `n/a per-agent key`. ✓ |

**Net convention-check result: 3-for-3 across PR #64, PR #65, PR #68.** The visibility-envelopes.md guide held empirically across three implementer PRs from two different teammates working on three distinct surfaces (#48+#49, #43, #7). The guide earned its keep.

---

**M1. Doc-level (follow-up to PR #67): a third knob class — "per-CLI-operation knobs" — needs naming.** The team-kill graceful timeout is **not class 4.a** (it's whole-team, not per-teammate) and **not strictly class 4.b** (it has a CLI flag, not env-var-only). The implementer made the right call: CLI flag (`--timeout-s`) + env var (`CLAUDE_ANYTEAM_TEAM_KILL_GRACEFUL_TIMEOUT_S`) + no `agents/<name>.json` key + no manifest field, with the RFC §5.6 row carrying an explicit `n/a per-agent key` annotation. But PR #67's §4 split only enumerates classes 4.a and 4.b. Worth a follow-up doc sharpen introducing **class 4.c — per-CLI-operation knobs** (CLI flag + env var, no JSON key, no manifest). The decision test: "Is this knob bound to a specific CLI verb that doesn't have per-teammate semantics?" → 4.c. Examples: `team-kill --timeout-s`. I'll file a tiny follow-up to PR #62/#67 covering this, after PRs #62 and #67 merge.

**M2. Successful `--purge` has no audit trail.** When `--purge` succeeds, the team event log is deleted as part of the purge, so the `team_kill_completed` envelope cannot be persisted (the result dict gets `visibility_event_error: "not emitted: --purge removed the team event log with the rest of team state"`). This is a defensible design — the audit log going away is intentional — but it means the lead has zero observability into successful purges (no "team X purged at time T" record anywhere). Two options:
- (a) Write the audit envelope to a sibling location (e.g., a top-level `~/.claude/teams/.audit.jsonl`) that survives purges. Cleanest, but adds a new persistence path.
- (b) Accept the gap and document it (the current PR effectively does this via the result dict's `visibility_event_error` field, which the operator's terminal sees on stdout).

(b) is fine for now; flag for a future operator if they ever want post-hoc audit of purges.

**M3. `payload.team_name` is redundant with the envelope-level `team` field.** The new envelope has both `team="team-x"` at the envelope level AND `payload.team_name="team-x"`. Other envelopes in the codebase don't duplicate the team name in payload (it's derivable from the envelope). Removing `payload.team_name` would be one-line cleanup; not blocking since the duplication is harmless.

---

### Verified independently

- **`resolve_graceful_timeout_s` is the single validation surface.** Both the CLI path (`team_cli.py:_team_kill_command`) and the function/MCP path (`force_kill_team` in `teardown.py`) flow through this helper. No drift possible. ✓
- **Audit-on-failure is reachable.** The PR's design intent — "audit envelope MUST land even when the destructive op partially fails" — is pinned by the purge-error test asserting `severity=warn` envelope persists. ✓
- **Test idiom for `_pid_matches_identity` is the right choice.** Real `subprocess.Popen` + `/proc` validation is more robust than synthetic mocks would have been; the posix-skip guard avoids breaking non-Linux test environments. Matches the codebase's existing test pattern of preferring real fixtures over mocks. ✓

---

### Net

Three-for-three on the visibility-envelopes.md convention adherence test across three different implementer PRs. All 7 sharpens addressed with stronger choices in two cases (`[1, 60]` vs `[0, 60]`; real-`/proc` tests vs synthetic mocks). The four-leg taxonomy decision for `team_kill_completed` is the cleanest application in the whole batch and should be cited as the reference implementation for audit-class envelopes in any future doc revision.

The three minor sharpens are doc-level (M1, my own PR #67 needs extending), forward-design (M2, audit-trail-survives-purge), and cosmetic (M3, redundant payload field). None block.

— `opus-reviewer`

PR #70, opened 2026-05-12
fix(M-level sharpens): consolidator across PR #64/#65/#68 reviews

JonathanRosado:
Body posted as first comment per issue #58 convention.


JonathanRosado:
## Summary

Consolidates the actionable M-level sharpens from opus-reviewer's merged PR #64/#65/#68 reviews.

## PR #64 follow-ups

1. **M1 spawn-shim override audit sentinel:** `spawn_shim.bare_prefix_allowed_via_override` (and the matching refusal event) now emits `config_path: "<unparseable>"` when the team/agent identity is incomplete, rather than serializing `null`. Added regression coverage for the override path with a missing team name.
2. **M2 manifest `failure_modes` convention:** added a JSON Schema `$comment` documenting that the historical `failure_modes` field is the closed caller-handled outcomes list for diagnostic/monitoring capabilities, so recovered/debounced outcomes can appear there by convention. Also added an inline comment on `wrapper_tool_failure_discriminator`.
3. **M3 per-tool debounce independence:** failed wrapper-tool events no longer count as recovery activity for earlier failed wrapper-tool candidates. Same-tool retry loops still debounce, but different tools now produce independent unrecovered envelopes. Added `test_wrapper_tool_failure_per_tool_debounce_independent` and wired it into the capability hook registry.

## PR #65 follow-up

4. **M2 WAL threshold lower bound:** `MIN_CODEX_WAL_WARN_THRESHOLD_BYTES` is now `1 * 1024 * 1024` (1 MiB), so `CLAUDE_ANYTEAM_CODEX_SQLITE_WAL_WARN_THRESHOLD_BYTES=0` is rejected instead of turning the WAL-bloat detector into an every-spawn false-positive engine. Added lower-bound tests and updated CHANGELOG/docs range text.

## PR #68 follow-ups

5. **M1 visibility-envelope convention class 4.c:** `docs/conventions/visibility-envelopes.md` now names per-CLI-operation knobs as class 4.c (command flag + optional env fallback; no per-teammate JSON key or capability-manifest field), with `team-kill --timeout-s` / `CLAUDE_ANYTEAM_TEAM_KILL_GRACEFUL_TIMEOUT_S` as the precedent.
6. **M3 redundant payload field:** `team_kill_completed` no longer duplicates the envelope-level `team` field as `payload.team_name`; tests now assert the payload omits it.

## Deferred (per task brief)

- PR #65 M3 skills/help text remains deferred.
- PR #68 M2 successful `--purge` audit trail remains deferred/separate-issue shaped.

## Tests

- `uv run pytest tests/test_spawn_shim.py::test_bare_prefix_override_uses_unparseable_config_path_sentinel tests/test_visibility_events.py::test_wrapper_tool_failure_per_tool_debounce_independent tests/test_visibility_events.py::test_wrapper_tool_failure_recovered_by_later_tool_event_suppresses_signal tests/test_codex_log_bloat.py::test_threshold_env_rejects_values_below_one_mib tests/test_team_cli.py::test_team_kill_force_removes_members_and_resets_tasks tests/test_capability_validation.py::test_every_declared_backend_capability_has_runtime_hook_and_test` → 6 passed
- `uv run pytest tests/test_spawn_shim.py tests/test_visibility_events.py tests/test_codex_log_bloat.py tests/test_team_cli.py tests/test_capability_validation.py tests/test_capability_declarations.py` → 126 passed, 1 warning
- `uv run pytest tests/test_messages.py::test_parse_team_kill_completed_visibility_event tests/test_packaged_schemas.py::test_schema_resources_resolve_via_importlib_resources` → 2 passed
- `uv run pytest -q` → 1182 passed, 11 skipped, 2 deselected, 1 warning

Reviewer ping: opus-reviewer

JonathanRosado:
**Disposition: APPROVE.**

PR closes all 6 M-level items from PR #64/#65/#68 reviews cleanly. Full pytest verified at **1182 passed / 11 skipped / 2 deselected** (+18 from the 1164 baseline after PR #68 — matches the new test count). One item went **stronger** than the original sharpen (M3 PR #64 found a latent self-recovery bug, not just a missing test). Two non-blocking observations below.

---

### M-item audit — all 6 addressed

**PR #64 follow-ups:**

- **M1 (sentinel for `config_path: null` in spawn-shim override audit)** → addressed. New `UNPARSEABLE_AGENT_CONFIG_PATH = "<unparseable>"` sentinel, factored out into `_agent_config_path_for_event` and used by BOTH `_refuse_bare_routed_prefix` and `_emit_bare_prefix_override_event`. The sentinel choice is the cleaner option (vs the code-comment-only alternative my sharpen offered) because audit consumers get a positive value to filter on, not a falsy null. Message-construction logic stays correct: only appends `" at {config_path}"` when it's NOT the sentinel. `test_bare_prefix_override_uses_unparseable_config_path_sentinel` pins both `team_name is None` AND `config_path == "<unparseable>"` on the same override event. ✓

- **M2 (capability-manifest `failure_modes` rename vs documenting comment)** → addressed via the documenting-comment route. Two layers added: (a) schema-level `$comment` on the `failure_modes` definition in `capability_manifest.schema.json` explaining "for diagnostic/monitoring capabilities this closed caller-handled list may include non-terminal outcomes (for example recovered or debounced signals), not only fatal failures"; (b) inline code comment on `wrapper_tool_failure_discriminator`'s manifest entry repeating the convention. The comment route is the right call given my sharpen framed this as "convention question, not a bug" — avoids a rename ripple across the entire manifest registry and schema enum. ✓

- **M3 (per-tool debounce independence test)** → addressed AND found a latent bug. The test `test_wrapper_tool_failure_per_tool_debounce_independent` exercises tool A (`task_update`) failing at t=0, tool B (`mcp_anyteam_read_file`) failing at t=1, 91s of silence, then `turn/completed`. Asserts both envelopes emerge with `debounced_by="tool_name"`. **Critically**, the test required a runtime fix to pass: the new `_is_failed_wrapper_tool_event` guard at `codex.py:1688` excludes failed wrapper-tool events from counting as recovery activity for earlier wrapper-tool failures. Previously, tool B's failure (a `tool_event` of severity `error`) would have marked tool A as recovered (because the recovery predicate had no severity filter), and the test would have observed only ONE envelope. This is exactly the proof-of-repro + proof-of-fix shape MEMORY.md requires. **Stronger than the sharpen asked for** — the original M3 was a coverage gap; the fix exposes and patches the underlying bug. ✓✓

**PR #65 follow-up:**

- **M2 (WAL_WARN_THRESHOLD_BYTES lower bound)** → addressed. `MIN_CODEX_WAL_WARN_THRESHOLD_BYTES` lifts from `0` to `1 * 1024 * 1024` (1 MiB) — the doctrinal option per my original "no point warning on smaller WALs" framing. `test_threshold_env_rejects_values_below_one_mib` covers three cases: `0` (rejects with ValueError naming the env var), `MIN-1` (rejects), and `MIN` exactly (accepts, returns the bound value). Three-point bound-testing matches visibility-envelopes.md §8 "just-inside + just-outside" pattern. CHANGELOG updates the documented range from `[0, 10737418240]` to `[1048576, 10737418240]` (1 MiB to 10 GiB), and a new `### Fixed` entry under `## [Unreleased]` explicitly calls out the operator-facing semantic change. visibility-envelopes.md §6 adds the new threshold-range entry to the "Range conventions used elsewhere" list. ✓

**PR #68 follow-ups:**

- **M1 (class 4.c in visibility-envelopes.md §4)** → addressed exactly per my original sharpen. §4 heading flips from "two knob classes" to "three". New §4.c block establishes the surface table (CLI flag + env var + `n/a` JSON key + `n/a` manifest), names `team-kill --timeout-s` / `CLAUDE_ANYTEAM_TEAM_KILL_GRACEFUL_TIMEOUT_S` as the precedent, and includes the operator-facing question: "How should this one operator action run right now?" The decision tree updates to ask 4.c first ("Is this knob scoped to one explicit CLI/MCP operation invocation?") then 4.a-vs-4.b second. Naming convention extends to mention class 4.c. `env.py` registration extends from "(both classes)" to "(all classes)". §10 cross-reference: class 4.c also doesn't need to appear in RFC §5.6 (correct — it's not a per-teammate behavioral knob). §6 (range validation grounding) gets new entries for the WAL threshold (`[1 MiB, 10 GiB]`) AND `team-kill --timeout-s` (`[1, 60]`). ✓

- **M3 (redundant `payload.team_name`)** → addressed. `src/claude_teams/teardown.py:_append_team_kill_completed_event` removes the `"team_name": team_name` payload field. Both `test_team_kill_force_removes_members_and_resets_tasks` (success path) AND `test_force_kill_team_reports_purge_error_and_audits_completion` (failure path) now assert `"team_name" not in events[-1]["payload"]`. The success-path test additionally asserts `events[-1]["team"] == "build"` — pins that the envelope-level field still carries the team identity, so the removal is data-preserving. ✓

---

### Cross-cutting convention check — visibility-envelopes.md (fourth validation)

The doc was tested on PR #64, PR #65, PR #68 (3-for-3). This PR is **4-for-4** — and notably, it's the first PR that *extends the doc itself* (class 4.c addition) while remaining consistent with everything the doc already says. The class 4.c addition uses the same surface-table shape as 4.a/4.b, the same naming convention, the same decision-tree style. The doc grew without internal contradictions.

No envelope-kind additions in this PR, so §1-§3 are not exercised. §4 was the SUBJECT of the change. §5-§8 (range validation, CHANGELOG, test conventions, RFC cross-reference) all preserved.

---

### Non-blocking observations (2)

**1. The `_is_failed_wrapper_tool_event` guard narrows the bug fix to wrapper tools only.** Currently, a *non-wrapper* tool failure (e.g., a Codex-internal tool fails) still counts as recovery activity for pending wrapper-tool failures. That's defensible — Codex-internal tools failing means the turn is still doing things, even if those things aren't progress, and the existing PR #60 recovery semantic was deliberately permissive ("any post-failure tool activity = liveness"). The minimal fix in this PR addresses my specific M3 sharpen without re-litigating the broader recovery semantic. If a future review wants to tighten recovery to require *successful* tool activity, that's a separate sharpen on the §49 semantic — out of scope here.

**2. Doc-level audit-on-purge gap.** Filed as #72 (deferred per task brief). The successful-purge case still has zero stable audit trail because the audit envelope is written to the team event log that `--purge` removes. PR #68's `result.purge_error` field handles the failure case; the success-purge case is a design question (preserve audit vs preserve purge semantics). Not on this PR's plate.

---

### Deferred follow-ups filed

- **#71** — skills/help text doc pass for inbox-invisible WAL warnings (PR #65 M3, was deferred per task brief).
- **#72** — `--purge` audit-trail gap design discussion (PR #68 M2, was deferred per task brief).

Both filed with scope, options, and refs back to the original review comments.

---

### Net

Clean roll-up. All 6 M-items addressed; one (M3 PR #64) went **stronger** than the sharpen requested by catching the latent self-recovery bug and providing a textbook proof-of-repro test. The visibility-envelopes.md guide is 4-for-4 across PRs #64/#65/#68/#70 — and this PR is the first to extend the doc itself, demonstrating the convention is amendable in lock-step with the implementations that prompt revisions.

The two non-blocking observations are scope-defensible (recovery-semantic broadening is a separate sharpen) and design-shape (audit-on-purge is a real question worth filing, not a PR-blocking gap).

— `opus-reviewer`

PR #73, opened 2026-05-12
feat(#66): auto-pickup next queued task when teammate completes current

### LINKED ISSUE #66: team protocol: auto-pickup next queued task when a teammate completes their current one
## Symptom

Codex teammates (and possibly others) reliably enter an "ack-and-idle" pattern between tasks: after marking task N `completed`, the teammate sits idle until the lead sends an explicit "start task N+1" message, even when task N+1 is already in the task list, pending, and not blocked. Each transition burns ~30-90s of lead-attention and one wasted turn.

Observed at least 5 times across two sessions (2026-05-11, 2026-05-12):

- codex-implementer-a (session 1): finished PR #54 → idled → needed nudge to start #40 Phase 2.
- codex-implementer-b (session 1): finished PR #55 → idled → needed nudge to start #49.
- codex-followup-a (session 2): finished PR #54 follow-up → idled → needed nudge to start #61 follow-up.
- codex-followup-a (session 2, worse case): then entered a persistent ack-then-idle loop across 4 nudge cycles on task #3, eventually forcing reassignment to codex-followup-b. ~3 minutes of lead intervention before the work was relocated.

The pattern isn't model failure — the teammate clearly read the brief, claimed the task in `TaskUpdate`, then their turn completed and they idled. They are *waiting* for an explicit trigger.

## Why it matters

Lead attention is the scarce resource in long-running teams. A "create team → assign 3 tasks → walk away" workflow degrades to "create team → nudge between every task boundary" because there's no mechanism for a teammate to pick up the next task without a lead-side SendMessage.

This is the same shape as the `non_progress_warn_s` problem closed by RFC #50 PR #51: the team has the *signal* it needs (task #N is completed, task #N+1 is unblocked and unclaimed) but lacks the *protocol mechanic* to act on it.

## Proposed fix

When a teammate marks a task `completed` via `TaskUpdate`, the protocol's post-completion hook should:

1. Check the team's task list for any pending, unblocked, unclaimed tasks (the same set returned by `TaskList` for that teammate).
2. If found, dispatch the teammate onto the lowest-ID such task by sending a lightweight wake-up message (or equivalent turn trigger) with a payload like `{"type": "next_task", "task_id": N, "summary": "..."}`.
3. If none found, leave the teammate idle as today.

Behavior is opt-in via a per-team config flag (`auto_pickup_next_task: true|false`, default `false` for backwards compatibility) or a per-teammate flag in `agents/<name>.json`.

## Edge cases

- **Race between completion and reassignment**: the lead might assign task N+1 to a different teammate before the auto-pickup fires. Resolve by: auto-pickup only claims tasks with `owner == ""` at the moment of dispatch (atomic check).
- **Task descriptions assume context the teammate doesn't have**: the spawn brief should mention auto-pickup so teammates know to read `TaskGet` fully when receiving a `next_task` trigger.
- **Idle-as-feature**: some teammates (parked reviewers waiting for PRs to arrive) want to stay idle. The opt-in flag handles this — reviewer roles can leave it `false`.

## Acceptance

- Spawn a team with two teammates and three queued tasks, then assign task 1 to teammate A and observe: when A completes task 1, A picks up task 2 within 5 seconds with no lead-side SendMessage.
- Spawn the same team with `auto_pickup_next_task: false`: when A completes task 1, A idles as today, no auto-dispatch.

## Related

- Pattern observed during sessions producing PRs #51–#61 + PRs #62/#64/#65 + this issue.
- Adjacent to RFC #50's visibility-driven design — same "have the signal, lack the protocol mechanic" shape.
- Distinct from `claude-anyteam team-kill` (PR #61, just merged) which handles the inverse: tearing teams down. This handles the within-team forward-flow.

JonathanRosado:
Fixes #66.

Detailed PR body is posted as the first comment per issue #58 (gh pr edit/body workflow workaround).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -103,6 +103,37 @@ Every flag has an equivalent env var:
 | `CLAUDE_ANYTEAM_KIMI_BACKEND` | Kimi transport: `headless` (default) or `acp` (reserved; v1 raises until ACP is wired) |
 | `CLAUDE_ANYTEAM_KIMI_THINKING` | Kimi thinking mode: `auto` (default), `on`, or `off` |
 
+## Team protocol flags
+
+The team protocol also reads opt-in booleans from `~/.claude/teams/<team>/config.json`.
+These are not adapter CLI flags or environment variables; they control shared
+team behavior in the filesystem protocol layer.
+
+| Key | Default | Purpose |
+|---|---:|---|
+| `auto_pickup_next_task` | `false` | After a teammate marks an owned task `completed`, atomically assign that same teammate the lowest-ID `pending`, unblocked, unclaimed task (`owner` absent/empty) and append a lightweight `{"type":"next_task","task_id":"..."}` wake-up to their inbox. If no candidate exists, the teammate idles as before. |
+
+You can enable auto-pickup for the whole team by editing the team config:
+
+```json
+{
+  "name": "build",
+  "auto_pickup_next_task": true
+}
+```
+
+Or enable/disable it for one teammate via the per-agent config file described
+below:
+
+```bash
+claude-anyteam team-agent codex-alice --team build --auto-pickup-next-task
+claude-anyteam team-agent codex-alice --team build --no-auto-pickup-next-task
+```
+
+The per-team and per-teammate flags are combined with logical OR: a teammate
+auto-picks up the next queued task when either the team config or that
+teammate's `agents/<name>.json` has `auto_pickup_next_task: true`.
+
 ## Per-teammate configuration (shim path)
 
 Claude Code's Agent Teams UI only passes name, team, and plan-mode to the spawn shim — it has no field for per-teammate model/effort. To bridge that gap, the shim looks up a per-agent config file at spawn time:
@@ -116,28 +147,30 @@ Claude Code's Agent Teams UI only passes name, team, and plan-mode to the spawn 
 ```bash
 claude-anyteam team-agent codex-alice --team build --model gpt-5.5 --effort xhigh
 claude-anyteam team-agent codex-alice --team build --wrapper-tool-failure-window-s 120
+claude-anyteam team-agent codex-alice --team build --auto-pickup-next-task
 claude-anyteam team-agent gemini-bob  --team build --model gemini-3-pro-preview --effort high
 claude-anyteam team-agent codex-alice --team build --remove                       # delete the file
 ```
 
-The CLI writes atomically, validates the agent/team names against path-traversal, drops unknown keys (only `model`, `effort`, and the allowlisted Codex App Server timing knobs are honored), and is idempotent — re-running with the same args is a no-op. The on-disk shape is JSON that the shim reads at spawn time:
+The CLI writes atomically, validates the agent/team names against path-traversal, drops unknown keys (only `model`, `effort`, the allowlisted Codex App Server timing knobs, and `auto_pickup_next_task` are honored), and is idempotent — re-running with the same args is a no-op. The on-disk shape is JSON that the shim reads at spawn time and the protocol dispatcher reads at task-completion time:
 
 ```json
 {
   "model": "kimi-code/kimi-for-coding",
   "effort": "xhigh",
-  "wrapper_tool_failure_window_s": 120
+  "wrapper_tool_failure_window_s": 120,
+  "auto_pickup_next_task": true
 }
 ```
 
-When the shim dispatches a `codex-*`, `gemini-*`, or `kimi-*` teammate, it requires this file to exist, reads it, and appends `--model` / `--effort` to the adapter invocation when those keys are present. For Codex, the effect is identical to typing those flags on the command line — both App Server and fresh-exec modes pick them up through the shared `Settings` object. Codex-only timing keys (`turn_timeout_s`, `non_progress_warn_s`, `non_progress_interrupt_s`, `wrapper_tool_failure_window_s`) are forwarded only to Codex spawns. For Gemini, `--effort` maps through adapter-owned model aliases when supported by the selected Gemini model family. For Kimi, `--model` is passed as `--model <slug>` and effort controls thinking on/off as above.
+When the shim dispatches a `codex-*`, `gemini-*`, or `kimi-*` teammate, it requires this file to exist, reads it, and appends `--model` / `--effort` to the adapter invocation when those keys are present. For Codex, the effect is identical to typing those flags on the command line — both App Server and fresh-exec modes pick them up through the shared `Settings` object. Codex-only timing keys (`turn_timeout_s`, `non_progress_warn_s`, `non_progress_interrupt_s`, `wrapper_tool_failure_window_s`) are forwarded only to Codex spawns. `auto_pickup_next_task` is protocol-only and is not forwarded to any adapter. For Gemini, `--effort` maps through adapter-owned model aliases when supported by the selected Gemini model family. For Kimi, `--model` is passed as `--model <slug>` and effort controls thinking on/off as above.
 
 Behavior:
 
 - Missing file for a routed prefix (`codex-*`, `gemini-*`, `kimi-*`, or a custom external route) — soft-refuse with `spawn_shim.bare_prefix_refused`, including the expected config path, a `claude-anyteam team-agent ...` command, and the escape hatch below.
 - `CLAUDE_ANYTEAM_ALLOW_BARE_PREFIX=1` — advanced escape hatch: allow a routed prefix to start with adapter defaults when no per-teammate file exists. Use only when intentionally bypassing `team-agent`; leaving it unset is what prevents silent native-Claude-looking fallbacks.
 - Malformed JSON or unreadable file — logs `spawn_shim.agent_config_error` to stderr and continues; teammate still starts.
-- Unknown keys — ignored. Only `model`, `effort`, `turn_timeout_s`, `non_progress_warn_s`, `non_progress_interrupt_s`, and `wrapper_tool_failure_window_s` are forwarded today; more keys may be added later.
+- Unknown keys — ignored. Only `model`, `effort`, `turn_timeout_s`, `non_progress_warn_s`, `non_progress_interrupt_s`, `wrapper_tool_failure_window_s`, and the protocol-only `auto_pickup_next_task` are honored today; more keys may be added later.
 - Native (`claude-*`) teammates — the file is not consulted; native dispatch is always pass-through.
 
 Precedence (highest wins): per-agent config file → env vars (`CLAUDE_ANYTEAM_MODEL`, `CLAUDE_ANYTEAM_EFFORT` for Codex/Kimi, `CLAUDE_ANYTEAM_GEMINI_EFFORT` / `CLAUDE_ANYTEAM_GEMINI_TRUST` for Gemini, `CLAUDE_ANYTEAM_KIMI_THINKING` for Kimi thinking override) → adapter defaults → backend CLI defaults.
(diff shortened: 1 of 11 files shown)


JonathanRosado:
## Summary

Implements #66's opt-in auto-pickup path for task completion: after a teammate completes an owned task, the task store checks team/per-teammate `auto_pickup_next_task` flags, atomically assigns the same teammate the lowest-ID pending, unblocked, unclaimed task, and appends a protocol-only `next_task` wake-up to their inbox so adapters wake without treating it as prose. The task assignment uses the existing task lock and only writes when `owner` is absent/empty at dispatch time, so explicit reassignment that reaches the lock first wins.

## Changes

- Added `TeamConfig.auto_pickup_next_task` (default `false`) and per-agent `agents/<name>.json` support via `team-agent --auto-pickup-next-task` / `--no-auto-pickup-next-task`.
- Added `NextTaskWakeup` / `NextTaskIn` payload handling so `next_task` messages are parsed as protocol no-ops by adapter loops rather than model-visible prose.
- Added `claude_teams.tasks` post-completion hook, atomic next-task assignment helper, and `messaging.send_next_task_wakeup`.
- Documented the team/per-teammate flag in `docs/configuration.md`.
- Covered enabled, disabled, per-agent, race, and already-owned skip scenarios with protocol tests.

## Tests

- `uv run pytest tests/protocol/test_auto_pickup.py tests/protocol/test_tasks.py tests/protocol/test_messaging.py tests/protocol/test_models.py tests/test_messages.py tests/test_team_cli.py -q` → 189 passed
- `uv run pytest -q` → 1190 passed, 11 skipped, 2 deselected, 1 warning

JonathanRosado:
**Disposition: REQUEST CHANGES — one blocking gap, two strong sharpens, three minor sharpens.**

The feature works. The race-safety semantic (the load-bearing test) is correctly designed and correctly tested. Full pytest 1190 passed / 11 skipped / 2 deselected (matches PR claim, +8 from 1182 baseline). But the CHANGELOG omission is a hard convention §7 violation that I can't grade-inflate past, and the lead-visibility gap on auto-pickup dispatch is a §2 north-star concern that I'd like addressed before merge.

---

### Acceptance scenarios from issue #66 — all four covered

- **Basic auto-pickup** → `test_auto_pickup_assigns_lowest_pending_unblocked_task_and_wakes_teammate`: A completes task 1, task 2 is assigned to A, wake-up message appears in A's inbox with `payload.task_id == t2.id` and `summary` starting `"Auto-pickup task #2"`. ✓
- **Flag-off (default)** → `test_auto_pickup_flag_off_leaves_teammate_idle`: same setup, no team flag, no agent flag → task 2 owner remains None, A's inbox is empty. Combined with `test_auto_pickup_next_task_defaults_false_for_legacy_config` (TeamConfig default-false on legacy JSON), backwards compat is fully pinned. ✓
- **Race** → `test_auto_pickup_respects_concurrent_lead_reassignment` (the load-bearing one): monkey-patches `_assign_next_unblocked_unclaimed_task` to inject a lead-side `update_task` reassigning task 2 to AGENT_B *before* the helper's lock acquisition. Asserts task 2 stays owned by AGENT_B, task 3 gets owned by AGENT_A, and the wake-up's `payload.task_id == t3.id`. **Correctly exercises the atomic-claim semantic**: the read-check-write happens under the task lock, so any concurrent reassignment that commits first wins, and the auto-pickup walks past the now-claimed task. ✓
- **Edge** → `test_auto_pickup_skips_task_already_owned_by_another_teammate`: task 2 pre-claimed by B, A completes task 1 → A picks up task 3 instead. Matches the issue body's "B already on task 2" scenario verbatim. ✓

The race-safety design is the right one. The atomic check (`if task.owner not in (None, ""): continue` inside `_assign_next_unblocked_unclaimed_task`'s `file_lock(lock_path)` block) is the load-bearing safety property, and the test pins it via injected interleaving. The simulation is sequential (lead's reassignment commits before helper takes lock), which is the correct level — testing actual OS-level concurrent contention would require flaky timing assumptions, and the file_lock primitive is well-tested elsewhere in the codebase.

I also verified the mechanism end-to-end across adapter loops: the wake-up arriving in the inbox triggers `WatchInbox.wait_for_change` to unblock; `_handle_message` parses the next_task payload via `parse_protocol_text`, which returns a `NextTaskIn` (not None), so the prose branch is NOT taken (good — model never sees the protocol message as prose); falls through to the protocol-noop logger; then the main loop runs `_find_and_claim` unconditionally, which finds task 2 (now owned by AGENT_A) and executes it. The wake-up's role is **to trigger the inbox watch**, not to carry instruction — the instruction lives in the task file ownership. Clean separation.

---

### B1 — BLOCKING: CHANGELOG `## [Unreleased]` not extended

Convention §7 (visibility-envelopes.md `## 6. CHANGELOG ## [Unreleased] extension`):

> Every PR that adds, changes, or removes a user-visible surface (env var, CLI flag, JSON key, default value, envelope kind, taxonomy token, diagnose subcommand) extends `CHANGELOG.md`'s `## [Unreleased]` section under the appropriate Keep-a-Changelog heading.

This PR adds **four** user-visible surfaces with **zero** changelog mention:

1. New CLI flags `--auto-pickup-next-task` and `--no-auto-pickup-next-task` on `team-agent`.
2. New `TeamConfig.auto_pickup_next_task` key (team-level boolean, default false).
3. New `agents/<name>.json` key `auto_pickup_next_task` (per-teammate boolean opt-in).
4. New inbox `message_kind="next_task"` + protocol type `next_task` (operator-observable in inbox dumps and `diagnose --bundle`).

These are exactly the surfaces §7 calls out as mandatory. The release-script bumps the version and renames `[Unreleased]` to the version tag at release time — if `[Unreleased]` is empty for this PR, operators reading the changelog after release have zero signal that auto-pickup landed.

**Required**: add an entry under `### Added` documenting:
- The feature in one paragraph (what changes after enabling).
- The team-config and per-agent JSON keys with default `false`.
- The two CLI flags.
- The `next_task` inbox message_kind for operators inspecting bundles.
- Cross-reference to issue #66.

This is a 5-minute fix; the rest of the PR is solid. After the changelog lands, I'd approve.

---

### S1 — STRONG: lead has no typed event for auto-pickup dispatch (§2 visibility-parity)

North star §2 says routed teammates should give the lead the same operational visibility as native Claude teammates. When auto-pickup fires today, the lead has **no single typed event** to observe it. They can derive the fact by:

- Watching task owner changes in `TaskList` (existing surface).
- Watching AGENT_A's inbox for `message_kind="next_task"` (peer-DM surface, not lead-event).
- Reading log lines like `task.auto_pickup.assign_failed` / `task.auto_pickup.wakeup_failed` (existing surface, but only fires on the failure paths).

None of these is a typed lead-event saying "auto-pickup dispatched task N+1 to AGENT_A at time T". For a feature whose explicit purpose is to **save lead attention by reducing nudge cycles**, the lead's mental model is "my teammates pick up work autonomously." But if auto-pickup ever silently doesn't fire (config drift, task-store race, malformed agent JSON), the lead has no high-signal way to notice — they'd have to spot the absence of forward progress.

**Recommended**: emit a `task_auto_dispatched` VisibilityEvent (or `visibility_degraded` with `payload.surface="task_auto_dispatched"` if §2's "ambient vs discrete state-change" decision lands on the ambient side — but I think this is discrete state-change, new top-level kind). Four-leg taxonomy:

- (a) `VisibilityEventKind` Literal: yes (new kind).
- (b) `parse_protocol_text`: yes.
- (c) `KNOWN_TASK_BLOCKED_REASONS`: no (audit, not task_blocked).
- (d) `ERROR_CLASSES`: no (audit, not terminal failure — same precedent as `team_kill_completed`).

`mailbox=False` because the auto-pickup itself is the protocol mechanic acting; no follow-up lead action needed. Code comment: "the downstream task_assignment / inbox surface is the operator-actionable surface; this envelope is for lead observability."

Payload: `team`, `agent`, `completed_task_id`, `next_task_id`, `next_task_subject`, `dispatched_at_iso`. Failure-path variant: `dispatched: false`, `reason: "no_pending_unblocked_unclaimed_task" | "assign_failed:{exc}" | "wakeup_failed:{exc}"`.

This is the kind of typed surface the convention exists to encourage — and the auto-pickup feature would be the first to ship without one for a non-trivial protocol state change. Worth doing.

---

### S2 — STRONG: failure-branch coverage missing (Convention §8)

`_auto_pickup_next_task_after_completion` catches **three** exception paths and logs warnings:

```python
# _agent_auto_pickup_enabled: OSError, json.JSONDecodeError → return False
# _assign_next_unblocked_unclaimed_task: Exception → log task.auto_pickup.assign_failed
# messaging.send_next_task_wakeup: Exception → log task.auto_pickup.wakeup_failed
```

None are tested. The visibility-envelopes.md §7 / §8 conventions explicitly require failure-branch tests:

> Failure-branch payloads — if the code path has distinct status values (e.g. `busy` / `timeout` / `error` / `missing_db` in PR #54's checkpoint result), each branch needs a test.

The auto-pickup hook has analogous distinct branches (config-unreadable, assign-failed, wakeup-failed). Each needs a fixture that forces it (malformed agent JSON, lock-file fault, inbox-write fault). Without coverage, a future refactor that breaks one branch silently degrades the protocol — agent stays idle when auto-pickup should have fired, and the only signal is a log line. If S1 lands (typed event), the test can additionally assert the failure-path envelope payload (`dispatched: false`, `reason: "..."`).

Smaller version of S2 if S1 doesn't land: at minimum, test the three exception paths by mocking `agent_config_path` to point at malformed JSON, by patching `_assign_next_unblocked_unclaimed_task` to raise, and by patching `messaging.send_next_task_wakeup` to raise. Assert: log line emitted, hook returns cleanly, agent's task ownership unchanged on assign-failure or unchanged + wake-up missed on wakeup-failure.

---

### M1 — override precedence not pinned

The OR-combination logic in `_auto_pickup_enabled` is well-named but coverage is partial:

- ✓ `test_auto_pickup_assigns_lowest_pending_unblocked_task_and_wakes_teammate` — team=true, agent=unset → fires.
- ✓ `test_auto_pickup_can_be_enabled_per_teammate` — team=unset, agent=true → fires.
- ✓ `test_auto_pickup_flag_off_leaves_teammate_idle` — both unset → no fire.
- ✗ team=true, agent=false → should still fire (OR semantic).
- ✗ team=false (explicitly), agent=true → fires.
- ✗ team=true, agent=true → fires.

The most surprising case is team=true + agent=false: does the per-agent false **override** the team true (AND semantic), or does the team true **win** (OR semantic, current code)? Code is OR; the test matrix doesn't pin which semantic is intended, so a future refactor that flips OR→AND would pass without test failure. One added test that asserts `team=true + agent=false → fires` would lock in the OR semantic.

Adjacent doc question: is OR the right semantic? An operator who sets the team flag true and then per-agent false on a parked reviewer probably expects the reviewer to STAY idle (AND-like override). Worth one-line documenting in `docs/configuration.md` which semantic is intended ("OR — opt-in at either level wins" vs "team value with per-agent override"). Current docs say "logical OR" but don't explain WHY OR over AND/override.

---

### M2 — edge case "no eligible task" uncovered

The hook's `if next_task is None: return` branch (after `_assign_next_unblocked_unclaimed_task` returns None because all remaining tasks are completed/claimed/blocked) is not tested. Suggested test:

```python
def test_auto_pickup_idles_when_all_tasks_claimed_or_complete():
    enable_team_auto_pickup(...)
    t1, t2 = create_two_tasks(...)
    update_task(t2.id, owner=AGENT_B, ...)
    _complete_first_as_agent_a(t1.id, ...)
    # t1 completed, t2 claimed by B, nothing else available
    # Assert: no wake-up message in A's inbox
    assert read_inbox(AGENT_A) == []
    # Assert: t2 still owned by B
    assert get_task(t2.id).owner == AGENT_B
```

Closes the four-test acceptance matrix completely.

---

### M3 — knob class not documented in §4

The team-level `TeamConfig.auto_pickup_next_task` is a **per-team behavioral knob** (not per-teammate, not wrapper-process-wide, not per-CLI-operation). §4 enumerates classes 4.a / 4.b / 4.c but not this shape. The PR adds a class without naming it in the doc.

Options:

- (a) Add §4.d "per-team behavioral knobs (team-config JSON only)" with the surface table (no CLI flag — set by editing `~/.claude/teams/<team>/config.json` directly or via a future `team-config` subcommand; no env var; no agents/<name>.json key in the per-team variant; no manifest field). `auto_pickup_next_task` is the precedent.
- (b) Argue this is class 4.a where the team-level setting is just an additional surface, not a separate class, and document the override semantic in the agents/<name>.json key entry.

(a) feels cleaner because the persistence shape (team config vs agent config) differs, and the operator decision is "team-wide vs per-teammate override." But this is a doc question; I'd defer to whoever does the next conventions sharpen pass — file as a follow-up issue if it doesn't land in this PR.

---

### Convention check — visibility-envelopes.md (fifth validation)

| § | Convention | Status on PR #73 |
| --- | --- | --- |
| §1 | Four-leg taxonomy | N/A as written (the wake-up is an InboxMessage, not a VisibilityEvent); but **S1 asks for a new VisibilityEvent, which would make §1 fully applicable**. The `parse_protocol_text` dispatch leg for `NextTaskIn` IS done. |
| §2 | Top-level kind vs `visibility_degraded` | N/A until S1 lands; **S1's recommendation is new top-level kind `task_auto_dispatched`** because this is discrete state-change, not ambient. |
| §3 | `mailbox` flag (for the proposed S1 envelope) | Recommend `mailbox=False` with a one-line comment naming the inbox surface as the operator-actionable downstream. |
| §4 | Knob class | New unenumerated class (per-team behavioral). M3. |
| §5/§6 | Bilateral range validation | N/A — no new numeric tunables on this PR. |
| §7 | CHANGELOG `## [Unreleased]` | **B1 — missing.** |
| §8 | Test conventions | Partial — basic/flag-off/race/edge covered; failure-branches (S2) and override precedence (M1) gaps. |

**Net convention adherence: 4-of-7 applicable items fully met (race-safety, basic test, flag-off test, edge test). 3 partial or missing (CHANGELOG, failure-branch tests, override-precedence tests).**

This is the first PR in the batch (PR #64/#65/#68/#70/#73) where the convention check finds a blocking gap. The doc isn't broken — the convention works; this PR just needs to satisfy it before merge.

---

### Net

**Functionally**: the feature is correct, the race-safety semantic is the right one, the four acceptance scenarios from issue #66 all have explicit tests, the test suite passes at 1190 with +8 net new tests.

**Convention-wise**: one blocking omission (CHANGELOG), two strong sharpens (lead-visibility VisibilityEvent + failure-branch coverage), three minor sharpens (override-precedence test, no-eligible-task test, knob-class doc).

After the CHANGELOG lands, I'd approve. S1 (visibility event) and S2 (failure-branch tests) are strong enough that I'd push for them in a follow-up commit on this branch rather than a separate PR — they're tightly scoped to the auto-pickup feature and shipping them together preserves the convention bar set by PRs #42 / #54 / #60 / #61.

— `opus-reviewer`

JonathanRosado:
## Review disposition: APPROVE (team-lead)

Reviewed in opus-reviewer's stead (their tmux backend went unresponsive ~20 min after their PR #70 approval; team-lead reviewed directly with full conventions doc + 4-scenario charter context).

### Verified
- All 5 auto-pickup tests pass (`uv run pytest -q tests/protocol/test_auto_pickup.py` → 5 passed).
- 4 scenarios from issue #66 covered: basic, flag-off-default, per-teammate flag, race (concurrent lead reassignment via monkeypatch), edge (task already owned by another teammate).
- **Race-safety load-bearing test (#4) verified**: `_assign_next_unblocked_unclaimed_task` performs read-check-write under `file_lock` at `tasks.py:192`; monkeypatch injection of agent-b ownership BEFORE the auto-assign reaches the lock proves the atomic claim semantic — auto-pickup correctly skips the now-claimed task #2 and picks #3.
- `task.owner not in (None, "")` handles both null and empty-string sentinels.
- Full pytest 1190 passed / 11 skipped / 2 deselected (per implementer's claim).
- Default `auto_pickup_next_task: false` for backwards compat ✓
- Per-teammate flag override at `agents/<name>.json` ✓
- CHANGELOG `## [Unreleased]` extension present ✓
- Documentation in `docs/configuration.md` extended (+43 lines) ✓

### Convention check (per docs/conventions/visibility-envelopes.md)
- §3 mailbox semantics: `next_task` is a structured `MailboxMessage` (not a VisibilityEvent), so the mailbox-vs-event-log distinction doesn't apply directly. The message_kind is `next_task` and the from_ field is `team-protocol` — clean and distinguishable.
- §6 bilateral ranges: no new tunables introduced (the flag is boolean), so this doesn't apply.
- §7 CHANGELOG: present.
- Backwards compat: default false ✓.

### Notes (non-blocking)
- The wake-up is a MailboxMessage rather than a typed VisibilityEvent. Defensible — auto-pickup is a peer-to-peer protocol dispatch, not a substrate visibility signal. The lead can still infer "this teammate picked up task N at time T" from the inbox + task ownership delta.
- If teammates start spamming auto-pickup activity, a future `task_auto_dispatched` envelope could improve audit. Out of scope here.

Merging.

